PR #11, opened 2022-06-27
Feature/release workflow

DanielGrauel:
Generating Container and Release Workflow


.github/workflows/dbc2val_feeder.yml
```diff
@@ -0,0 +1,76 @@
...
+# * information regarding copyright ownership.
+# *
+# * This program and the accompanying materials are made available under the
+# * terms of the Apache License 2.0 which is available at
+# * http://www.apache.org/licenses/LICENSE-2.0
+# *
+# * SPDX-License-Identifier: Apache-2.0
+# ********************************************************************************/
+
+name: val-feeder
```
SebastianSchildt, line 14: @argerus  can we rename the worklfows to dbc-feeder, or kuksa-dbc-feeder , becasue basically all components here are "Val feeders", I suggest `kuksa_dbc_feeder` , because that fits the GPS feeder naming

SebastianSchildt (changes requested):

.github/workflows/dbc2val_feeder.yml
argerus: Done

PR #12, opened 2022-07-20
Change log level to error

DanielGrauel:
Change the log level from debug to error cause of Task 10824


SebastianSchildt (approved):
Looking fine

Issue #14, opened 2022-08-08
Update GPS Feeder to VSS 3.0


SebastianSchildt:
Together with
https://github.com/eclipse/kuksa.val/issues/322

GPS feeder uses hardcodes VSS locations still in Infotainment. Adapt it to feed GPS to Vehicle.CurrentLocation

erikbosch:
Fixed by #17 - can be closed

PR #17, opened 2022-08-18
Update to VSS 3.0 syntax

### LINKED ISSUE #13: Update CAN mapping to VSS V3.0
Together with https://github.com/eclipse/kuksa.val/issues/322

Once the new default VSS model in kuksa.val is VSS 3.0, we should update the CAN feeder mapping accordingly

erikbosch:
Depends on VSS 3.0 support in kuksa.val
Fixes #13 and #14
Draft commit, ready for review but not fully tested yet.

### DIFF
--- a/dbc2val/mapping.yml
+++ b/dbc2val/mapping.yml
@@ -16,7 +16,7 @@
 #
 # There is a lot more to be found in the traces. Feel free to suggest PRs
 #
-# This maps to VSS 2.2 
+# This maps to VSS 3.0
 # 
 # Uses model 3 DBC by Josh Wardell  https://github.com/joshwardell/model3dbc
 #
@@ -27,7 +27,7 @@
 # Note VSS is angle of the steering wheel, DBC might be angle or steering?
 # Converting to int
 SteeringAngle129:
-  vss: # taken from
+  vss: # taken from https://github.com/COVESA/vehicle_signal_specification/blob/master/spec/Chassis/Chassis.vspec
     datatype: int16 
     type: sensor
     unit: degrees
@@ -67,7 +67,7 @@ DI_gear:
     changetype: 2 # ON_CHANGE number taken from types_pb2.py
   minupdatedelay: 100
   targets:
-    Vehicle.Powertrain.Transmission.Gear:
+    Vehicle.Powertrain.Transmission.CurrentGear:
       transform:
         fullmapping:
           DI_GEAR_D: 1
@@ -81,13 +81,13 @@ DI_gear:
     datatype: bool
     type: sensor
     unit: none
-    description: The current gear. 0=Neutral, 1/2/..=Forward, -1/-2/..=Reverse
+    description: Is the transmission park lock engaged or not. False = Disengaged. True = Engaged.
   databroker:
     datatype:  1 # BOOL number taken from types_pb2.py
     changetype: 2 # ON_CHANGE number taken from types_pb2.py
   minupdatedelay: 100
   targets:
-    Vehicle.Chassis.ParkingBrake.IsEngaged:
+    Vehicle.Powertrain.Transmission.IsParkLockEngaged:
       transform:
         fullmapping:
           DI_GEAR_D: "false"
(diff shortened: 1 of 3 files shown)


SebastianSchildt:
@erikbosch If it is draft I think you can just mark/change it as draft in GH. Github Actions  can still be run

gps2val/gpsd_feeder.py
```diff
@@ -54,11 +54,11 @@ def shutdown(self):
         self.client.stop()
 
     def setPosition(self, position):
-        self.client.setValue('Vehicle.Cabin.Infotainment.Navigation.CurrentLocation.Altitude', position['alt'])
-        self.client.setValue('Vehicle.Cabin.Infotainment.Navigation.CurrentLocation.Latitude', position["lat"])
-        self.client.setValue('Vehicle.Cabin.Infotainment.Navigation.CurrentLocation.Longitude', position["lon"])
-        self.client.setValue('Vehicle.Cabin.Infotainment.Navigation.CurrentLocation.Accuracy', position["hdop"])
-        self.client.setValue('Vehicle.Cabin.Infotainment.Navigation.CurrentLocation.Speed', position["speed"])
+        self.client.setValue('Vehicle.CurrentLocation.Altitude', str(position['alt']))
```
erikbosch, line 57: For me running gps2val failed without adding `str` when using gpsfake and the included log file as input, it was complaining that `float`was not supported. I have tested with `str` towards kuksa-val-server and it works as expected.

@SebastianSchildt - do we know of anyone using this file that could help verifying this change.

erikbosch:
Following scenarios now tested with VSS 3.0 update

- val-server with viss-client
- val-server with dbc2val
- val-server with gps2val
- val-server unittest
- databroker with cli
- databroker with dbc2val

gps2val/gpsd_feeder.py
SebastianSchildt: Ah, that is looking fine. Has to do with some typing assumptions that changed in the python SDK.

SebastianSchildt:
@erikbosch GPS feeder tested and works for me, but now I modified it again, to support some more VSS 3.0 datapoints if they are provided by gpsd. Can you check, if it still works as expected on your side?

erikbosch:
@SebastianSchildt - new commit added, fixing the `FluidLevelLow` error. Also adding another mapping to avoid some warnings. (But we still get a lot of other errors as described in #18)

dbc2val/mapping.yml
```diff
@@ -411,6 +412,7 @@ VCLEFT_rearDoorState:
         fullmapping:
           DOOR_STATE_CLOSED: "false"
           DOOR_STATE_OPEN_OR_AJAR: "true"
+          DOOR_STATE_RELEASING_LATCH: "true"
```
erikbosch, line 415: @SebastianSchildt - we got warnings on that this value was not handled. I added a line to also consider that value. As this is an example mapping I do not know if exact mapping is that important. As I understand this is an intermediate state so we could theoretically handle it in 3 ways:

- Consider that the door is opened when this is reported
- Consider that it is still closed
- Ignore this state, but then we should rather have a syntax so that we can state that explicitly so we do not get a warning.
Any comments?

SebastianSchildt (changes requested):
Still not mapped correctly

```
2022-08-25 09:57:30,228 ERROR dbcfeeder: Error setting Vehicle.Powertrain.ElectricMotor.Motor.Torque: I can not find Vehicle.Powertrain.ElectricMotor.Motor.Torque in my db
```

dbc2val/mapping.yml
SebastianSchildt: For the example mapping such details are not important, so the solution in this mapping is fine.  A clean solution might be to have  variant of the mapping transform.  Currently there is "fullmapping", which, as you have noticed will print a warnming if something can not be append, and we have partialmapping, that will just pass through the raw value, if no mapping is found (i.e. this can be used if you only want to remap a subset of states), what we are missing is a mapping transform  that maps, what it can, and silently ignores non-mapped values. Future work :)

SebastianSchildt (approved):
Works for me.
Let's revisit existing databroker issues after API unification

PR #21, opened 2022-08-26
Use same condition for ghcr upload and close

erikbosch:
To avoid failed builds if there are no tokens. Previously most builds failed, see https://github.com/eclipse/kuksa.val.feeders/actions/workflows/clean-images.yml

```
Run bots-house/ghcr-delete-image-action@v1.0.0
  with:
    owner: eclipse
    name: kuksa.val.feeders/gps
    tag: pr-17
  env:
    REGISTRY: ghcr.io
    GPSIMAGE_NAME: eclipse/kuksa.val.feeders/gps
Error: Input required and not supplied: token
```

Tested that ignoring cleaning works if there is no token in: https://github.com/eclipse/kuksa.val.feeders/actions/runs/2933358224

### DIFF
--- a/.github/workflows/clean-images.yml
+++ b/.github/workflows/clean-images.yml
@@ -10,11 +10,29 @@ env:
   GPSIMAGE_NAME: ${{ github.repository }}/gps
 
 jobs:
+  check_tokens:
+    runs-on: ubuntu-latest
+    outputs:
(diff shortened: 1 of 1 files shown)


SebastianSchildt (approved):
Let's merge and see how it behaves down the road

PR #24, opened 2022-09-29
Upgrade docker build environment for gps2val

erikbosch:
Background:
gps2val relies on kuksa-viss-client. For Alpine docker env there is no prebuilt grpcio for aarch64 so the build action need to build it itself. The grpcio wheel does not seem to include all dependencies required, like linux-headers.

Now aligning Dockerfile with https://github.com/eclipse/kuksa.val/blob/master/kuksa_viss_client/Dockerfile.
See also https://github.com/eclipse/kuksa.val/issues/367 for general discussions on grpc/arrach64/Alpine


SebastianSchildt (approved):
lgtm

PR #25, opened 2022-10-20
Feature/someip feeder

int0x27:
SOME/IP to KUKSA.VAL DataBroker feeder

- Uses vsomeip library to subscribe and receive an example "Wiper" Service events.
- Provides vsomeip example wiper service for local (on single host, unix sockets only) and remote (udp) testing.
- Provides helpers for KUKSA.VAL Databroker integration.
- Provides generic someip client for receiving SOME/IP Events.
- Example someip configurations and scripts for modifying those configurations.

**NOTE:** dockerisation of someip feeder is not ready, only generic someip examples are dockerised as POC


someip2val/docker/vsomeip/examples/notify-sample.cpp
```diff
@@ -0,0 +1,305 @@
+// Copyright (C) 2014-2017 Bayerische Motoren Werke Aktiengesellschaft (BMW AG)
```
SebastianSchildt, line 1: We need an Eclipse Apache header here if we argue the example is rather documentation or we need to do an CQ for MPL if we consider this a modified someip ocde

someip2val/docker/vsomeip/examples/subscribe-sample.cpp
```diff
@@ -0,0 +1,207 @@
+// Copyright (C) 2014-2017 Bayerische Motoren Werke Aktiengesellschaft (BMW AG)
```
SebastianSchildt, line 1: same as above.

Also in any case we need to add Eclipse KUKSA (c) above to not give the impression BMW is responsible for whatever bad things we might have done to their code

lukasmittag:
Works for me!

someip2val/docker/vsomeip/examples/notify-sample.cpp
int0x27: I've removed modified examples, they had some dumps added, nothing important

SebastianSchildt (approved):
lgtm

PR #27, opened 2022-11-03
mapping.yml: Remove mapping to multiple targets (Vehicle.Speed and Vehicle.OBD.Speed)

BjoernAtBosch:


dbc2val/mapping.yml
```diff
@@ -53,7 +53,8 @@ DI_uiSpeed:
   minupdatedelay: 100
   targets: 
     Vehicle.Speed: {}
-    Vehicle.OBD.Speed: {}
+    # Commented out, because mapping to multiple targets is not supported currently
```
SebastianSchildt, line 56: I think it is fine to comment out the OBD variant (as Vehicle.Speed is the "better" datapoint),

However, as I think some older examples (node.red) still depend on Vehicle.OBD.Speed, can you make the comment more precise and say
 
```
# Commented, because mapping to multiple targets currently not supported in databroker
```

I think then it is an easy fix for now. Once we migrate this feeder to the updated Python Lib all the "custom code" goes out anyway, and then we will make this work again

SebastianSchildt (changes requested):
Generally fine, but see comment

dbc2val/mapping.yml
BjoernAtBosch: Done

SebastianSchildt (approved):
lgtm

PR #35, opened 2022-11-29
Lukasmittag/update gps feeder

lukasmittag:
Using a library for a lightweight gpsd client isntead of the provided client library of the gpsd project, because this client gets more frequently released as pip package. Also this ports to the use of the new python library which is compliant either to KUKSA.val server or databaroker.

### DIFF
--- a/gps2val/gpsd_feeder.py
+++ b/gps2val/gpsd_feeder.py
@@ -34,7 +34,7 @@ from gpsdclient import GPSDClient
 
 scriptDir= os.path.dirname(os.path.realpath(__file__))
 sys.path.append(os.path.join(scriptDir, "../../"))
-from kuksa_viss_client import KuksaClientThread
+from kuksa_client import KuksaClientThread
 
 class Kuksa_Client():
 
@@ -47,8 +47,9 @@ class Kuksa_Client():
         provider_config=config['kuksa_val']
         self.client = KuksaClientThread(provider_config)
         self.client.start()
-        print("authorizing...")
-        self.client.authorize()
+        if str(provider_config.get('protocol')) == 'ws':
+            print("authorizing...")
+            self.client.authorize(str(provider_config.get('token')))
         
     def shutdown(self):
         self.client.stop()
--- a/gps2val/gpsd_feeder.py
+++ b/gps2val/gpsd_feeder.py
@@ -31,6 +31,7 @@ import csv
 import time
 import queue
 from gpsdclient import GPSDClient
+import argparse
 
 scriptDir= os.path.dirname(os.path.realpath(__file__))
 sys.path.append(os.path.join(scriptDir, "../../"))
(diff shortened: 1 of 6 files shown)


gps2val/gpsd_feeder.py
```diff
@@ -59,7 +59,7 @@ def setData(self, data):
             if v is not None:
                 self.client.setValue(k,str(v))
                
-class GPSD_Client():
+class GPSD_Client_Instance():
```
romainletendart, line 63: Having "instance" in a class name is confusing as an object is an instance of a class already.
Also, it seems to me this class could be a subclass of `threading.Thread` and that the `loop` method could be replaced with an override of https://docs.python.org/3.8/library/threading.html?highlight=thread#threading.Thread.run.
We could then rename the class `GPSDClientThread` (without underscores which are forbidden in class names).

gps2val/gpsd_feeder.py
```diff
@@ -83,37 +83,32 @@ def __init__(self, config, consumer):
...
+            if self.running:
+                print("")
+                self.collecteddata['Vehicle.CurrentLocation.Latitude']= result.get('lat',None)
+                self.collecteddata['Vehicle.CurrentLocation.Longitude']= result.get('lon',None)
+                self.collecteddata['Vehicle.CurrentLocation.Altitude']= result.get('alt', None)
+                self.collecteddata['Vehicle.Speed']= result.get('speed',None)
+                self.collecteddata['Vehicle.CurrentLocation.Timestamp']= result.get('time',None)
+                self.collecteddata['Vehicle.CurrentLocation.Heading']= result.get('track',None)
+                self.collecteddata['Vehicle.CurrentLocation.HorizontalAccuracy']= result.get('eph',None)
+                self.collecteddata['Vehicle.CurrentLocation.VerticalAccuracy']= result.get('epv',None)
```
romainletendart, line 97: You don't need to specify `,None` as it's already the default value.

gps2val/gpsd_feeder.py
```diff
@@ -83,37 +83,32 @@ def __init__(self, config, consumer):
...
+                break
 
      
 
 
     def shutdown(self):
-        self.running=False
-        self.consumer.shutdown()
-        self.gpsd.close()
+        print("KUKSA client shutdown")
```
romainletendart, line 108: I don't see where you shut down the kuksa client.
I think you need to resurrect the removed `self.consumer.shutdown()` line.

lukasmittag: Must've happened while rebase should'nt be deleted. You're right.

gps2val/requirements.txt
```diff
@@ -1,2 +1,2 @@
 kuksa-viss-client
```
SebastianSchildt, line 1: This should be kuksa-client now

gps2val/readme.md
```diff
@@ -2,10 +2,8 @@
 consumes [gpsd](https://gpsd.gitlab.io/gpsd/) as datasource and pushes location to kuksa.val server.
 The [`gpsd_feeder.ini`](./config/gpsd_feeder.ini) contains `kuksa.val` and `gpsd` configuration.
 
-Before starting the gps feeder, you need start `kuksa.val` and `gpsd`:
+Before starting the gps feeder, you need to start `kuksa.val server/databroker` (for this have a look at [KUKSA.val server](../../kuksa-val-server/README.md) and [KUKSA.val databroker](../../kuksa_databroker/README.md)). You have to start an instance of `gpsd` by running:
```
SebastianSchildt, line 5: Maybe use absolute links, because if you did not check this out as submodule (or just clicking in Github Webif) you can not reach the doc, so I suggest
../../kuksa-val-server/README.md -> https://github.com/eclipse/kuksa.val/blob/master/kuksa-val-server/README.md etc.

gps2val/requirements.txt
lukasmittag: Is it already available under kuksa-client?

romainletendart: Yes! https://pypi.org/project/kuksa-client/

romainletendart (approved):

SebastianSchildt:
Starting the container does not work, even when binding the config dir as in Readme

```
Init kuksa client...
  File "/kuksa_gps_feeder/./gpsd_feeder.py", line 128, in <module>
    gpsd_client = GPSDClientThread(config, Kuksa_Client(config))
  File "/kuksa_gps_feeder/./gpsd_feeder.py", line 48, in __init__
    self.client = KuksaClientThread(provider_config)
  File "/usr/local/lib/python3.10/site-packages/kuksa_client/__init__.py", line 42, in __init__
    self.backend = cli_backend.Backend.from_config(config)
  File "/usr/local/lib/python3.10/site-packages/kuksa_client/cli_backend/__init__.py", line 39, in from_config
    protocol = config['protocol']
  File "/usr/local/lib/python3.10/configparser.py", line 1258, in __getitem__
    raise KeyError(key)
KeyError: 'protocol'
```

SebastianSchildt (changes requested):
Same happens when starting it in a virtual py environment

SebastianSchildt (approved):
Nice. 🧑‍🎤

PR #41, opened 2022-12-12
Show how overlays can be used

erikbosch:
Currently only for discussion. Seems that tooling change in vss-tools would be needed

### DIFF
--- a/dbc2val/overlay/README.md
+++ b/dbc2val/overlay/README.md
@@ -1,0 +1,260 @@
+# Using overlays as alternative mapping
+
+## Example 1 - Boolean
+
+```
+VCFRONT_passengerPresent:
+  vss: # taken from https://github.com/COVESA/vehicle_signal_specification/blob/master/spec/Cabin/SingleSeat.vspec
+    datatype: boolean 
+    type: sensor
+    unit: none
+    description: Does the seat have a passenger in it.
+  databroker:
+    datatype: 1 # BOOL number taken from types_pb2.py
+    changetype: 2 # ON_CHANGE number taken from types_pb2.py
+  minupdatedelay: 1000
+  targets:
+    Vehicle.Cabin.Seat.Row1.Pos2.IsOccupied:
+      transform:
+        fullmapping:
+          1: "true"
+          0: "false"
+```
+
+A functionally equal mapping could possibly be like this:
+
+```
+Vehicle.Cabin.Seat.Row1.Pos2.IsOccupied:
+  type: actuator
+  datatype: boolean
+  dbc: VCFRONT_passengerPresent
+  dbc_mapping: [{1:true},{0:false}]
+  dbc_changetype: ON_CHANGE
+  dbc_interval_ms: 1000
+```
+
+Which could (theoretically, see below,) be called like:
+
+```
+./vss-tools/vspec2json.py --no-uuid -e dbc,dbc_mapping,dbc_changetype,dbc_interval_ms ./spec/VehicleSignalSpecification.vspec -o overlay.vspec --json-pretty vss.json
+```
+
+This gives this resulting JSON:
+
+```
+                      "IsOccupied": {
+                        "datatype": "boolean",
+                        "dbc": "VCFRONT_passengerPresent",
+                        "dbc_changetype": "ON_CHANGE",
+                        "dbc_interval_ms": 1000,
+                        "dbc_mapping": [
+                          {
+                            "$file_name$": "overlay.vspec",
+                            "$line$": 5,
+                            "$name$": "1:true"
+                          },
+                          {
+                            "$file_name$": "overlay.vspec",
+                            "$line$": 5,
+                            "$name$": "0:false"
+                          }
+                        ],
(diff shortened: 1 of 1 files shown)


dbc2val/overlay/README.md
```diff
@@ -0,0 +1,219 @@
...
+      transform:
+        fullmapping:
+          1: "true"
+          0: "false"
+```
+
+A functionally equal mapping could possibly be like this:
+
+```
+Vehicle.Cabin.Seat.Row1.Pos2.IsOccupied:
```
erikbosch, line 27: @SebastianSchildt - as of today we cannot support this type of overlays, see https://github.com/COVESA/vss-tools/issues/201

dbc2val/overlay/README.md
```diff
@@ -0,0 +1,219 @@
...
+  type: actuator
+  datatype: boolean
+  dbc: VCFRONT_passengerPresent
+  dbc_mapping: [{1:true},{0:false}]
```
erikbosch, line 31: Well we could actually have something very similar to how it was before, like


```
  transform:
    fullmapping:
      "1": true
      "0": false
```

(Current VSS-tools cannot handle integers as keys, but if we include them in quotation marks it works well)

So in short - we can represent mapping as a tree structure - did not thought that was possible

dbc2val/overlay/README.md
```diff
@@ -0,0 +1,219 @@
...
+
+Notes:
+* As of today the VSS overlay mechanism cannot be used on expanded branches like `Vehicle.Cabin.Seat.Row1.Pos2.IsOccupied`.
+  Creating a syntax that that works on "unexpanded" nodes like `Vehicle.Cabin.Seat.IsOccupied` would mean more work,'
+  both for defining the syntax and also for using it within dbc2val feeder. VSS-Tools [issue](https://github.com/COVESA/vss-tools/issues/201) created for discussion.
+* The term `minupdatedelay` currently used might be misleading.
+  Rather than specifying a delay, or a minimum interval, it specifies a wanted interval (for CONTINUOUS).
+  I.e. if you specify 1000, you want to get an update about every 1000 ms, you do not want it every 2000 ms.
+  For `ON_CHANGE` ita minimum interval.
+* For mapping it is likely easiest having a single VSS attribute as VSS-tooling does not support more advanced structures.
```
erikbosch, line 70: As stated above, this is not correct

argerus (changes requested):

dbc2val/overlay/README.md
argerus: I added a separate review comment, but maybe a reply here fits better into the discussion..

I don't think it's a good idea to put what is essentially a "value" as a key like this. Keys are (usually) used to describe the corresponding value, which is one reason restrictions like "keys can only be strings" can make sense.

The mapping can be specified as something like this instead:
```json
  "mapping": [
    {
      "from": 0,
      "to": false
    },
    {
      "from": 1,
      "to": true
    }
  ]
```
which would avoid the issue. Since this uses the keys to describe the values, the direction of the mapping is more clear as well (I think).

SebastianSchildt:
```  
transform:
    fullmapping:
      "1": true
      "0": false
```

If this works (or the equivalent of johns proposal), by "accident" in vss-tools, the question really is, do we want to allow arbitrary structures under extended attributes from VSS point of view? It might become painful for tooling

If it is seen "just as a string", (that my have some internal yaml/json/whatever structure) it might be easier, even if maybe not really "elegant" in the computer science sense...

(Not sure this is better discussed VSS upstream...)

On the other hand _if_ structures (as in "YAML in the originating vspec") are acceptable, probably all should go under one extended attribute, i.e. `dbc`

```yaml
Vehicle.Cabin.Seat.Row1.Pos2.IsOccupied:
  type: sensor
  datatype: boolean
  dbc:
    signal: VCFRONT_passengerPresent
    transform:
      mapping:
        - from: 0
          to: false
        - from: 1
          to: true
```

dbc2val/overlay/README.md
```diff
@@ -0,0 +1,260 @@
...
+
+```
+Vehicle.Cabin.Seat.Row1.Pos2.IsOccupied:
+  type: actuator
+  datatype: boolean
+  dbc: VCFRONT_passengerPresent
+  dbc_mapping: [{1:true},{0:false}]
+  dbc_changetype: ON_CHANGE
+  dbc_interval_ms: 1000
+```
```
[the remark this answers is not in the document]
SebastianSchildt: ahah what John says. I commented before scrolling to here 😁

dbc2val/overlay/README.md
argerus: I think it would be really beneficial if everything that is not "reserved" for use in VSS, was just passed along unmodified (to be used by the intended downstream tools or applications).

For YAML and JSON, it seems to work "accidentally" already..

For other formats, it would probably need some type of "extended attributes" and a `Variant` type holding the values of them, but maybe it doesn't have to be supported by all output formats anyway..

erikbosch:
Closing this one, replaced by #52

PR #46, opened 2023-01-13
Fix quoting in check_push_rights workflow

SebastianSchildt:
Fixes a quoting issue in check_push_rights found by @int0x27

### DIFF
--- a/.github/workflows/check_push_rights.yml
+++ b/.github/workflows/check_push_rights.yml
@@ -17,14 +17,14 @@ jobs:
     - name: check GITHUB_TOKEN allows GHCR push access
       id: check-secrets
       run: |
-          if [[ ${{ github.event_name }} == "push"  ]] &&  [[ ${{ github.repository }} == "eclipse/kuksa.val.feeders"  ]]; then
+          if [[ "${{ github.event_name }}" == "push"  ]] &&  [[ "${{ github.repository }}" == "eclipse/kuksa.val.feeders"  ]]; then
             echo "We are pushing to kuksa.val.feeders upstream, so we should have rights"
             echo "have_secrets=true" >> $GITHUB_OUTPUT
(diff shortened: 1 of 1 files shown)


.github/workflows/check_push_rights.yml
```diff
@@ -17,14 +17,14 @@ jobs:
     - name: check GITHUB_TOKEN allows GHCR push access
       id: check-secrets
       run: |
-          if [[ ${{ github.event_name }} == "push"  ]] &&  [[ ${{ github.repository }} == "eclipse/kuksa.val.feeders"  ]]; then
+          if [[ "${{ github.event_name }}" == "push"  ]] &&  [[ "${{ github.repository }}" == "eclipse/kuksa.val.feeders"  ]]; then
```
int0x27, line 20: Actually I've checked that `[[` is safe to use with empty strings

SebastianSchildt: yeah, but I had an issue of one of the github variables evaluates to an empty string, see here in kuksa.val (using basically the same workflow) https://github.com/boschglobal/kuksa.val/actions/runs/3910675145/jobs/6683149741

This was a "workflow_dispatch" in a repo that is not upstream, and the Github Action replacement leads to this line

```
  if [[ workflow_dispatch == "pull_request"  ]] &&  [[  == "eclipse/kuksa.val"  ]]; then
```

(as it is is not a PR, there is no repo where it might come from, thus the second clause becomes invalid.

Quoting fixes that.

int0x27: I suggest using `[` with quoted variables, although it is ok as is.
The confusion comes that `[[` can handle empty environment variables, but `${{}}` is replaced as string

SebastianSchildt:
Solved meanwhile

PR #50, opened 2023-01-27
Fix Dockerfile in CAN Feeder

lukasmittag:
Set Patchelf to hardcoded version

### DIFF
--- a/dbc2val/Dockerfile
+++ b/dbc2val/Dockerfile
@@ -30,7 +30,7 @@ ENV PATH="/opt/venv/bin:$PATH"
 RUN /opt/venv/bin/python3 -m pip install --upgrade pip \
     && pip3 install --no-cache-dir -r requirements.txt
 
-RUN pip3 install wheel && pip3 install scons && pip3 install pyinstaller && pip3 install patchelf && pip3 install staticx
+RUN pip3 install wheel scons && pip3 install pyinstaller patchelf==0.17.0.0 staticx
 
 RUN pyinstaller --clean -F -s dbcfeeder.py
 #   --debug=imports
@@ -73,4 +73,4 @@ ENV VEHICLEDATABROKER_DAPR_APP_ID=vehicledatabroker
 
 ENV PYTHONUNBUFFERED=yes
 
-CMD ["./dbcfeeder-exe"]
+ENTRYPOINT ["./dbcfeeder-exe"]


dbc2val/Dockerfile
```diff
@@ -30,7 +30,7 @@ ENV PATH="/opt/venv/bin:$PATH"
 RUN /opt/venv/bin/python3 -m pip install --upgrade pip \
     && pip3 install --no-cache-dir -r requirements.txt
 
-RUN pip3 install wheel && pip3 install scons && pip3 install pyinstaller && pip3 install patchelf && pip3 install staticx
+RUN pip3 install wheel && pip3 install scons && pip3 install pyinstaller && pip3 install patchelf==0.17.0.0 && pip3 install staticx
```
erikbosch, line 33: I verified that i could build the docker container locally with this fix. But I would like to have some clue somewhere on why we need to fix it to this particular version. In a comment in this file, or in the commit or in the PR. If it is a known issue, possibly link to a place describing it

So that if someone looks at this line in 1 year can find out why it was needed and check if it is safe to remove the fixed version (which I assume we would need at some time in the future)

msrn:
Wouldn't it be better to define these pip packages in the `requiremens.txt` file?

erikbosch:
> Wouldn't it be better to define these pip packages in the `requirements.txt` file?
> 
It might depend on what we want to include in `requirements.txt` and  `requirements-dev.txt`, i.e. what is the purpose/scope of the files. `patchelf` is not needed for native builds, so does it qualify for the existing files. And if it is only needed for build, is `dev` a better location?

SebastianSchildt:
While at it, last line of file

```
CMD ["./dbcfeeder-exe"]
```

Should be changed to ENTRYPOINT (so you can give parameters to feeder when starting the container)

lukasmittag:
> > Wouldn't it be better to define these pip packages in the `requirements.txt` file?
> 
> It might depend on what we want to include in `requirements.txt` and `requirements-dev.txt`, i.e. what is the purpose/scope of the files. `patchelf` is not needed for native builds, so does it qualify for the existing files. And if it is only needed for build, is `dev` a better location?

In my opinion it does not belong in requirements.txt for sure. If it belongs to requirements-dev.txt depends on the scope of this file but I think it does not belong in either of this txt files as it is no "real" requirement. It is not necessarily nedded to build the code. Only if you want to use staticx and the generation of an executable.

lukasmittag:
> While at it, last line of file
> 
> ```
> CMD ["./dbcfeeder-exe"]
> ```
> 
> Should be changed to ENTRYPOINT (so you can give parameters to feeder when starting the container)

Schould then also be changed in https://github.com/eclipse/kuksa.val.services/pull/54 ?

dbc2val/Dockerfile
```diff
@@ -30,7 +30,7 @@ ENV PATH="/opt/venv/bin:$PATH"
 RUN /opt/venv/bin/python3 -m pip install --upgrade pip \
     && pip3 install --no-cache-dir -r requirements.txt
 
-RUN pip3 install wheel && pip3 install scons && pip3 install pyinstaller && pip3 install patchelf && pip3 install staticx
+RUN pip3 install wheel scons && pip3 install pyinstaller patchelf==0.17.0.0 staticx
```
msrn, line 33:
```suggestion
RUN pip3 install wheel scons pyinstaller patchelf==0.17.0.0 staticx
```

This was left here

lukasmittag: Yeah it is intended. Otherwise the arm build fails because pyinstaller and staticx can not find scons and wheel packages

erikbosch (approved):

SebastianSchildt (approved):
Probably there is room for improvement here. Maybe even revisiting if the (somewhat fragile) staticx thingy is really an advantage in a container, but for now let's take this working version over the broken one :)

PR #52, opened 2023-02-01
Use VSS as mapping format for dbcfeeder

erikbosch:
This is an implementation of #41 

Apart from changing the config there are also some general refactoring:

- Most python files moved to a `dbcfeederlib` folder. This is to simplify inclusion in unit testing
- Added some unit tests, it is still quite basic but at least some to verify syntax parsing
- Added a "on_change" config.

Testing fully performed:

- Sending signals to KUKSA.val Databroker (native and as Docker container)
- Sending signals to KUKSA.val Server (native and as Docker container)

Testing partially performed:

- elmcan: Verified that we reach and call the `createvcan.sh` script
- j1939: Exception observed, see #56


dbc2val/dbcfeeder.py
```diff
@@ -239,39 +253,31 @@ def _run(self):
...
-                        log.debug("Updating DataPoint(%s, %s)", target, value)
-                        if self._server_type is ServerType.KUKSA_DATABROKER:
+                vss_observation = self._can_queue.get(timeout=1)
+                vss_mapping = self._mapper.get_vss_mapping(vss_observation.dbc_name, vss_observation.vss_name)
+                value = vss_mapping.transform_value(vss_observation.raw_value)
+                if value is None:
+                    log.warning(f"Value ignored for  dbc {vss_observation.dbc_name} to VSS {vss_observation.vss_name},"
+                                f" from raw value {value} of type {type(value)}")
+                elif not vss_mapping.change_condition_fulfilled(value):
+                    log.info(f"Value condition not fulfilled for VSS {vss_observation.vss_name}, value {value}")
```
SebastianSchildt, line 263: have not tested it yet, but might the log level be too high? In the past we observed if just printing some message xxx times per second will slow down the feeder a lot/generate a lot of load. And just from looking at the code, this is not an error condition, but something that we might expect to encounter regularly. So maybe verbose?.

dbc2val/dbcfeederlib/dbc2vssmapper.py
```diff
@@ -0,0 +1,310 @@
...
+            if "mapping" in self.transform:
+                tmp = self.transform["mapping"]
+                # Assumed to be a list
+                for item in tmp:
+                    from_val = item["from"]
+                    if from_val == value:
+                        new_val = item["to"]
+                        vss_value = new_val
+                        break
+            else:
```
SebastianSchildt, line 137: Would it be more clean/future proof to really check there is a math transform? And in else print something like "there seem to be transforms I just don't understand them"

I am thinking that might be more robust, if we add further transform types in the future and overlays/config files/dbcfeeder implementations get mixed

dbc2val/mapping.md
```diff
@@ -0,0 +1,308 @@
...
+
+There are two methods for creating a mapping file. The first method is to manually add DBC
+information to an existing VSS file in JSON format. It is important that the VSS file is compatible
+with the VSS file used by KUKSA.val. One way to ensure this is to use the same JSON file for both
+KUKSA.val and DBC Feeder. If for example your KUKSA.val (kuksa-val-server or kuksa-databroker)
+instance uses one of the JSON files in the [KUKSA.val repository](https://github.com/eclipse/kuksa.val/tree/master/data/vss-core)
+then you can annotate that file and use the annotated file in both KUKSA.val and the feeder.
+
+Annotating an existing VSS JSON file has however some drawbacks. If the JSON file is regenerated
+to support a new VSS version then the annotations must be manually transferred to the new vSS JSON.
```
SebastianSchildt, line 38: type vSS->VSS

dbc2val/mapping.md
```diff
@@ -0,0 +1,308 @@
...
+
+Specifying DBC signal name is mandatory. It must correspond to a DBC signal name defined in a DBC file.
+By default the DBC feeder use the [Model3CAN.dbc](Model3CAN.dbc) example file.
+
+`interval_ms` and `on_change` are optional and control under which conditions a value shall be forwarded.
+The `interval_ms` value indicates the minimum interval between signals in milliseconds.
+The `on_change: True` argument specifies that the VSS signal only shall be sent if the DBC value has changed.
+If none of them are specified it corresponds to `interval_ms: 1000, on_change: False`.
+If only `on_change: True` is specified it corresponds to `interval_ms: 0, on_change: True`
+
```
SebastianSchildt, line 109: if on-change is true and an interval is given (not sure that is alway useful), how is the logic? I.e.

I open the door-> set open

Then the signal toggles millsions of times within one second, (interval_ms: 1000), so I assume it is not forwarded?
Then after the 1000ms IF the signal is coming again, is "on-change" based on the last reported or on the last observed value? Observed might be a problem because door->open=reported to VSS, door->closed not bcasue within timeout, and then will never be reported, if compare to last observed state.

Might be a corner case, but at least we should now - and preferably describe - what happens here

dbc2val/mapping.md
```diff
@@ -0,0 +1,308 @@
...
+  dbc:
+    signal: RearPower266
+    interval_ms: 4000
+    transform:
+      math: "floor(abs(x/5))"
+```
+
+Transformation may fail. Typical reasons may include that the DBC value is not numerical,
+or that the transformation fails on certain values like division by zero.
+If transformation fails the signal will be discarded.
```
SebastianSchildt, line 132: discarded->ignored

SebastianSchildt: Edit: I see the `discarded`wording is just often/consistently.  Will not mark every occurrence. can live with it :) I just would have written ignored

dbc2val/test/test_mapping_transform/test.json
```diff
@@ -0,0 +1,242 @@
...
+          "$file_name$": "test.vspec",
+          "$line$": 114,
+          "on_change": true,
+          "signal": "S5",
+          "transform": {
+            "$file_name$": "test.vspec",
+            "$line$": 117,
+            "mapping": [
+              {
+                "$file_name$": "test.vspec",
```
SebastianSchildt, line 16: Are the $file_name$ things an artifact from the vss-tools? It looks like those "scratchpad" data the parser attaches for giving better error messages. Do we consider it a bug upstream and will fix it there? Because in "vanilla" VSS vspec these things are cleaned up.

dbc2val/dbcfeeder.py
erikbosch: Could be reduced to debug.

For now if running on info-level you get typically one line every time a signal is sent to broker (and if here is value is discarded), which could be quite many. So the question is - how verbose do we want to be on info-level, do we expect info-level to be usable for "real deployments" where performance is important?

erikbosch: As of today default is INFO for dbcfeeder. I have no problems using DEBUG on recurring messages. But then , if you would be interested in knowing if any signals (at all) has been sent to databroker then you would need to change to DEBUG, and then you would get a lot of other messages as well.

dbc2val/dbcfeederlib/dbc2vssmapper.py
erikbosch: The idea is that `extract_verify_transform` shall do the static check that mappings are understood, so here only well defined mappings shall exist. But we could add an else-path here anyway. As long as we do not have too smart linters (that complains that the code is dead) it shall be ok.

dbc2val/mapping.md
erikbosch: I try to describe it a bit further down, will update it to make it clearer. The filtering is done in two phases. This partially depends on that we have a queue in between. I assume math-transform can be costly from performance perspective, and that is why we only do it time condition is fulfilled.

On_Change is always compared with last value sent to KUKSA.val. So theoretically, f you are very quick in opening/closing your door and you have an interval of 1000 ms then there is a risk that the opening not will be reported to KUKSA.val as we only "poll/sample" values every 1000 ms. 

* If there is a time condition the time of the observation is compared with the stored value.
  If the time difference is bigger than the explicitly or implictly defined interval
  the stored time for the VSS-DBC combination is updated
  and evaluation continue with the next step.
* The DBC value is then transformed to VSS value. If transformation fails the signal is discarded.
* After transformation, if there is a change condition, the stored value is compared with the
  new value. If they are equal the signal is discarded.

erikbosch:
PR updated. Based on our discussions on printouts I created a short status printout, also showing max queue size so far. this would give at least some indication if your feeder works as expected

```
2023-02-14 17:37:31,844 INFO dbcfeederlib.databroker: Vehicle.Powertrain.Transmission.IsParkLockEngaged was already registered with type BOOLEAN
2023-02-14 17:37:31,847 INFO dbcfeederlib.databroker: Vehicle.Trailer.IsConnected was already registered with type BOOLEAN
2023-02-14 17:37:31,945 INFO dbcfeeder: Starting to process CAN signals
2023-02-14 17:37:31,948 INFO dbcfeeder: Number of VSS messages sent so far: 1, queue max size: 10
2023-02-14 17:37:31,950 INFO dbcfeeder: Number of VSS messages sent so far: 2, queue max size: 10
2023-02-14 17:37:31,955 INFO dbcfeeder: Number of VSS messages sent so far: 4, queue max size: 10
2023-02-14 17:37:31,965 INFO dbcfeeder: Number of VSS messages sent so far: 8, queue max size: 10
2023-02-14 17:37:32,060 INFO dbcfeeder: Number of VSS messages sent so far: 16, queue max size: 10
2023-02-14 17:37:32,374 INFO dbcfeeder: Number of VSS messages sent so far: 32, queue max size: 10
2023-02-14 17:37:33,056 INFO dbcfeeder: Number of VSS messages sent so far: 64, queue max size: 10
2023-02-14 17:37:34,444 INFO dbcfeeder: Number of VSS messages sent so far: 128, queue max size: 10
2023-02-14 17:37:37,255 INFO dbcfeeder: Number of VSS messages sent so far: 256, queue max size: 10
2023-02-14 17:37:42,844 INFO dbcfeeder: Number of VSS messages sent so far: 512, queue max size: 10
2023-02-14 17:37:53,989 INFO dbcfeeder: Number of VSS messages sent so far: 1024, queue max size: 10
^C2023-02-14 17:38:07,494 INFO dbcfeeder: Received signal 2, stopping...
2023-02-14 17:38:07,494 INFO dbcfeeder: Shutting down...
```

SebastianSchildt:
Does not work on Python 3.8, needs 3.10 (did not test 3.9). Is this documented?

SebastianSchildt:
Default loglevel seems to be Info when run locally but warning in container. Should probably be the same? Info for both, now that info is less noisy?

erikbosch:
Our only statement on Python version is "Check that at least python version 3 is installed". In CI we use whatever Python-version is default on ubuntu-latest. Docker file use Python 3.9 so that 3.9 has been tested. I assume we likely can get it to run on Python 3.8, but we should better document what versions we intend to support. Python 3.11 is out - if we are to test all it will be a bit cumbersome to test all Python versions.

erikbosch:
There are some minor changes between Python 3.8 and 3.10, I sometimes experience them for vss-tools, where CI use Python 3.8 (and I use 3.10 when running locally). This [stackoverflow](https://stackoverflow.com/questions/59101121/type-hint-for-a-dict-gives-typeerror-type-object-is-not-subscriptable) describe the root cause also this time, it was easy to adapt to 3.8, but I will add something on supported versions.

SebastianSchildt (approved):
Works for me
🐘

PR #53, opened 2023-02-01
Reintegrate elm bridge

### LINKED ISSUE #49: ELM bridge in dbc
As of today we have [elm2canbridge.py](https://github.com/eclipse/kuksa.val.feeders/blob/main/dbc2val/elm2canbridge.py). It is not called from other Python files and no instructions on how to use it.

It appears that it previously was called [from feeder](https://github.com/eclipse/kuksa.val.feeders/blob/bfd1799d4871f1fadccafd050af0e41b0d9b583e/dbc2val/dbcfeeder.py#L63)

Things to do:

- Put back the old code
- Create instructions on how the elm-feature can be tested, so that it preferably can be tested before each (major) release

erikbosch:
Fixes #49 

This PR does two things:

- Reintroduces the elmcan support
- Improves error messages when elmcan cannot be started

_Note: I do not have any possibilities to actually test that the elmcan code works. But now it is at least possible to use!_

### DIFF
--- a/dbc2val/Readme.md
+++ b/dbc2val/Readme.md
@@ -55,6 +55,8 @@ $ pip install -r requirements.txt
 
 2. Start the can player
 
+If you want to simulate CAN traffic on vcan0 by replaying a can dump file you can use the canplayer.
+
 ```console
 $ ./createvcan.sh vcan0
 $ canplayer vcan0=elmcan -v -I candump.log -l i -g 1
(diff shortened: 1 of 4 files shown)


SebastianSchildt (changes requested):
```python
self.initelm(elm,cfg['canspeed'], cfg['canack'])
```

needs to be "speed" instead of canspeed, as that is what it is called in the config section

SebastianSchildt (approved):

PR #54, opened 2023-02-02
Added hithub context dump

int0x27:
Fixed error for undefined variables, replaces [PR-46](https://github.com/eclipse/kuksa.val.feeders/pull/46)

Signed-off-by: Ivan Kirchev <Ivan.Kirchev@bosch.com>

### DIFF
--- a/.github/workflows/check_push_rights.yml
+++ b/.github/workflows/check_push_rights.yml
@@ -5,27 +5,44 @@ on:
         description: "In possession of ghcr.io tokens?"
         value: ${{ jobs.check_push_rights.outputs.have_secrets }}
 
-
 jobs:
   check_push_rights:
     runs-on: ubuntu-latest
+
+    # # Permissions needed to push to ghcr.io. may not be granted on eclipse repo
+    # permissions:
+    #   contents: read
+    #   packages: write
+
     outputs:
       have_secrets: ${{ steps.check-secrets.outputs.have_secrets }}
 
     steps:
+
+    - name: Dump GitHub Context
+      env:
+        GITHUB_CONTEXT: ${{ toJson(github) }}
+      run: |
+        echo "$GITHUB_CONTEXT"
+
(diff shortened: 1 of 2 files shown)


.github/workflows/check_push_rights.yml
```diff
@@ -5,27 +5,44 @@ on:
...
             # (note that some/all workflows int his repo might still opt to no push PR builds to GHCR)
           fi
-          if [[ ${{ github.event_name }} == "pull_request"  ]] &&  [[ ${{ github.event.pull_request.head.repo.full_name }} == "eclipse/kuksa.val.feeders"  ]]; then
-              echo "We are an internal pull request , so we should have rights"
+          if [[ "${{ github.event_name }}" == "pull_request" ]] && [[ "${{ github.event.pull_request.head.repo.full_name }}" == eclipse/* ]]; then
+              echo "We are an internal pull request, so we should have rights"
+              echo "have_secrets=true" >> $GITHUB_OUTPUT
+              exit 0
+          fi
+          if [[ "${{ github.event_name }}" == "pull_request" ]] && [[ "${{ github.event.pull_request.head.repo.full_name }}" == boschglobal/* ]]; then
```
erikbosch, line 44: I am a bit uncertain if we should mention boschglobal in code in main repo as boschglobal is a downstream fork. Any opinions @SebastianSchildt?

But if we do, shall we then add a comment why, like:
`
 # For convenience downstream repositories in possession of a valid ghcr token are listed here `

int0x27: So far this is needed for debugging the environment and deciding for a better approach.
Main concern is `github.token` could leak (for the moment it seems filtered)

e.g. we can use
- `github.repository_owner != 'eclipse'`
- `github.event.repository.fork == 'false'`
- `github.secrets != 'None'`

The idea is you should be able to fork eclipse repo and release to ghcr of your own fork.

PR #55, opened 2023-02-09
gps feeder fixes and improvements

int0x27:
Fixed several issues in gps2val:
**NOTE:** command line arguments are broken, to be addressed in another PR.
- Slightly reduced size of the image (apk cahce removed)
- Workaround for a sighandler deadlock in gpsd_feeder, preventing image from stopping: thread.join(1) + os._exit(1)
- Disabled SIGQUIT handler in order to force an app crash in case sighandler hangs in the future
- Added some extra dumps for config
- Dockerfile uses `entrypoint.sh` script to allow custom cmdline arguments to gpsd_feeder.py
**NOTE:** gpsd is started as background job and it has dumps now, `entrypoint.sh` makes sure gpsd is terminated after python exits. Also the image has a hardcoded config file, so command line args are overriden by it.

Also fixes and improvements in workflow file (e.g. make it git clone friendly):
- Improved github context checks to guess if we should have rights to push on `ghcr.io`
- Now should be able to run on forked repos (without self-hosted arm64 runner).
- In forked builds, just `linux/amd64` platform is built as qemu build for arm64 needs ~3h to finish.
- Added nice summary for ephemeral `ttl.sh` images / `ghcr.io` tags in workflow summary

Signed-off-by: Ivan Kirchev <Ivan.Kirchev@bosch.com>

### DIFF
--- a/gps2val/Dockerfile
+++ b/gps2val/Dockerfile
@@ -13,7 +13,7 @@ LABEL org.opencontainers.image.source="https://github.com/eclipse/kuksa.val.feed
 LABEL org.opencontainers.image.authors="Wenwen.Chen@de.bosch.com"
 
 # alpine-sdk and linux-headers needed for Alpine to build grpcio for aarch64, not needed for platforms with prebuilt grpcio
-RUN apk update && apk add gpsd alpine-sdk linux-headers
+RUN apk add --no-cache gpsd alpine-sdk linux-headers
 ADD . /kuksa_gps_feeder
 WORKDIR /kuksa_gps_feeder
 RUN pip install --upgrade pip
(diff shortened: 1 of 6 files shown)


SebastianSchildt:
I tried making the resulting image smaller, the Dockerfile was quite sloppe anyway.

Does not solve the root cause for the 413 (probably this https://github.com/containerd/containerd/pull/7459 ) but maybe solves our problem while making the container image better as well. Fingers crossed

gps2val/Dockerfile
```diff
@@ -1,4 +1,4 @@
-# Copyright Robert Bosch GmbH, 2020. Part of the Eclipse Kuksa Project.
+# Copyright Robert Bosch GmbH, 2020-2023. Part of the Eclipse Kuksa Project.
```
erikbosch, line 1: @SebastianSchildt - this license is quite different compared to the [DBC Docker file](https://github.com/eclipse/kuksa.val.feeders/blob/main/dbc2val/Dockerfile). Historical reasons or mistake?

int0x27:
> 

I doubt it will actually work as gpsd (+dependencies) are not copied in runtime image

gps2val/Dockerfile
SebastianSchildt: Historical reasons, EDL is basically public domain, and historically we have used it for some config or build scripts, where we thought "copyrighting" them is too ridiculous. I.e. I have no issues if somebody copies snippets and tricks out of our Dockerfiles without any attribution. 
But probably, to mak scanners happy it would be better to just add Apache header and hoping people have enough "criminal energy" to copy helpful stuff anyway

SebastianSchildt (changes requested):
Much better, but typo in feeders

`ttl.sh/kuksa.val.feefers/gps-78e46794a7f119a9d413c8d44da100f3dc79dd3e:1h`

feefers->feeders

In summary it might be good to use the "long-form" with three backticks for the commnand, i.e.

```
docker run -it --rm  ttl.sh/kuksa.val.feefers/gps-78e46794a7f119a9d413c8d44da100f3dc79dd3e:1h 
```

because then, on mouse-over the browser gives you a small button to copy directly

SebastianSchildt (approved):

PR #61, opened 2023-02-28
Add DDS provider

lukasmittag:


lukasmittag:
@AkhilTThomas

lukasmittag:
tests are missing as of now. Will be added in another commit.

dds2val/tests/integration/feeder/test_ddsfeeder_int.py
```diff
@@ -0,0 +1,104 @@
...
+    vdb_client = VehicleDataBrokerClient()
+
+    await run_dds_feeder_and_send_dds_msg(
+        create_NavSatFix_dds_messages(["52.25,12.25,0"]),
+        "Nav_Sat_Fix",
+        sensor_msgs.msg.NavSatFix,
+    )
+
+    # Verify data received in broker is the same as send above
+    response = await vdb_client.GetDatapoints(["Vehicle.CurrentLocation.Latitude"])
```
erikbosch, line 68: Here we use a client from Velocitas and in another file we specify that we shall use version v0.7.1. I see Velocitas as a downstream project from kuksa so having a dependency to Velocitas is not optimal as I see it. 

We only seem to use the Velocitas client to read a datapoint, so could we possibly manage without it? Possibly by adding a "get" method to https://github.com/boschglobal/kuksa.val.feeders/blob/feature/DDS_feeder/dds2val/src/databroker.py

lukasmittag: Sure, good point.

dds2val/tests/integration/feeder/test_ddsfeeder_int.py
```diff
@@ -0,0 +1,104 @@
...
+    # Use sdv SDK without dapr:
+    Config().disable_dapr()
+
+    databroker = subprocess.Popen(['docker', 'run', '--net=host', 'ghcr.io/eclipse/kuksa.val/databroker:master'], stdin=subprocess.PIPE,
+                     stdout=subprocess.PIPE)
+    # if it needs to pull the image it builds long
+    time.sleep(120)
+
+    # prepare vehicle data broker
+    vdb_client = VehicleDataBrokerClient()
```
erikbosch, line 59: Now this will maybe be irrelevant if we remove the Velocitas client, but here we rely on default host/ip as defined in velocitas. Well that we also do on line 53 when starting databroker. It works but the question is if it would be wiser to state it explicitly for all.

(There is a slight difference, the SDV client use "localhost", ddsfeeder use "127.0.0.1")

dds2val/tests/integration/feeder/test_ddsfeeder_int.py
```diff
@@ -0,0 +1,104 @@
...
+async def test_ddsfeeder_start():
+    "Test case to verify data send on dds bus is received in broker"
+
+    # Use sdv SDK without dapr:
+    Config().disable_dapr()
+
+    databroker = subprocess.Popen(['docker', 'run', '--net=host', 'ghcr.io/eclipse/kuksa.val/databroker:master'], stdin=subprocess.PIPE,
+                     stdout=subprocess.PIPE)
+    # if it needs to pull the image it builds long
+    time.sleep(120)
```
erikbosch, line 56: I understand this line, but it is a bit annoying to wait when not needed. Do we have any other ideas, like checking every 5 seconds if the container is up and running?

dds2val/src/README.md
```diff
@@ -0,0 +1,24 @@
...
+
+1. ```python3 -m venv env && source env/bin/activate```
+2. ```pip install -r requirements-kml.txt```
+
+## How to run
+Choose from local build or contanerization via docker.
+These steps are necessary:
+1. Run an instance of databroker aka: ```docker run -it --rm --net=host ghcr.io/eclipse/kuksa.val/databroker:master```
+2. Start the KML replay with an active local python virtual environment: ```python3 dds_kmlreplay.py directions.kml```
+3. Start the DDS feeder with either: ```docker run --rm -it --net=host ddsfeeder:latest``` or with an active local python virtual environment: ```python3 ddsfeeder.py```
```
erikbosch, line 24: I tried to run both variants (built docker container and directly with python), got similar errors:
```

erik@debian3:~/kuksa.val.feeders/dds2val/src$ !4778
docker run --rm -it --net=host ddsfeeder:latest
  File "ddsfeeder.py", line 41, in <module>
ImportError: attempted relative import with no known parent package
[8] Failed to execute script 'ddsfeeder' due to unhandled exception!

```

dds2val/src/ddsfeeder.py
```diff
@@ -0,0 +1,249 @@
...
+# pylint: disable=unused-import
+import Vehicle  # noqa: F401
+from cyclonedds.core import Listener
+from cyclonedds.domain import DomainParticipant
+from cyclonedds.sub import DataReader
+from cyclonedds.topic import Topic
+from grpc import ChannelConnectivity
+from kuksa_client.grpc import VSSClient, VSSClientError
+
+from . import databroker, vss2ddsmapper  # noqa: F401
```
erikbosch, line 41: If changing this line to

`import databroker, vss2ddsmapper`

... then both my docker build and native run works

dds2val/__init__.py
```diff
@@ -0,0 +1 @@
+# makes dds2val importable
```
erikbosch, line 1: Do we want a newline?

dds2val/src/databroker.py
```diff
@@ -0,0 +1,107 @@
...
+            if len(metadata) == 1:
+                self._name_to_type[name] = metadata[name].data_type
+                log.info(
+                    "%s was already registered with type %s",
+                    name,
+                    metadata[name].data_type.name,
+                )
+                return True
+
+            return self._register(name, data_type, description)
```
erikbosch, line 65: This file is more or less the same as https://github.com/eclipse/kuksa.val.feeders/blob/main/dbc2val/dbcfeederlib/databroker.py (well actually the dbc2val version has been updated as we cannot register). 

I am not saying that we shall change it now as I think we want a bigger change, like having either this file as part of client code (in kuksa.val repo) or refactoring kuksa.val.feeders so that we get a "generic feeder module/package/lib" that can be used by all Python-based feeders. So maybe keep it as is and add it to a refactoring issue. There is anyway a need for refactoring in dbcfeeder as server/broker-specific code is not as hidden/encapsulated as I would like to.

dds2val/src/idls/NavSatStatus.idl
```diff
@@ -0,0 +1,38 @@
+/*
```
erikbosch, line 1: Do we know where these files comes from?
I am thinking - if we know it could be an idea to add a README where we state source/version, so that we can keep track if there are newer versions.

dds2val/src/idls/vss/setup.py
```diff
@@ -0,0 +1,4 @@
+import setuptools
```
erikbosch, line 1: Do we need a license header also for very small python files, well possibly even for empty ones like `__init__.py`? (FYI: @SebastianSchildt )

lukasmittag: These are no files which include intelectual property. So I assume they need no license headers like config files and so on.

dds2val/src/mapping.yml
```diff
@@ -0,0 +1,73 @@
+###############################################################################
+# Copyright (c) 2022 Robert Bosch GmbH
+#
+# All rights reserved. This configuration file is provided to you under the
+# terms and conditions of the Eclipse Distribution License v1.0 which
+# accompanies this distribution, and is available at
+# http://www.eclipse.org/org/documents/edl-v10.php
+#
+# SPDX-License-Identifier: Apache-2.0 OR BSD-3-Clause
```
erikbosch, line 9: This is the first time we use this SPDX-combination. Looking at https://www.eclipse.org/org/documents/edl-v10.php it only says BSD-3

lukasmittag: @SebastianSchildt what should we do. These are the licenses from the "real" author of this code. I did not change them.

erikbosch (changes requested):
what I have done:

- Reviewed the code
- Succeeded running unit tests
- Succeeded getting data sent to databroker (after changing import path as shown in comment)

Apart from the import (which either should be fixed, or someone else check that current setup works for them) the rest is minor items where not necessarily everything needs to be fixed.

dds2val/src/mapping.yml
SebastianSchildt: True, but I think it is fine. Question is always "should" you give configs a complex license, or maybe none at all? (that is why in "old EPL" times in KUKSA we used  EDL for stuff we did not deem "worthy" to be protected by EPL. 

The OR does not hurt here, question is, and it is acutally in line with the old dbc mapping: https://github.com/eclipse/kuksa.val.feeders/blob/v0.1.1/dbc2val/mapping.yml

However, to streamline this, I would suggest either removiong any license headers (as this is only config), but not sure how much license scanners like this
 or adding the normal apache 2.0 only (trusting people have enough criminal energy to "steal without attribution/given back" when needed)

dds2val/src/databroker.py
erikbosch: A little bit related to https://github.com/eclipse/kuksa.val.feeders/issues/62

dds2val/tests/integration/feeder/test_ddsfeeder_int.py
lukasmittag: Or maybe use kuksa client library here. Will change

dds2val/src/ddsfeeder.py
```diff
@@ -0,0 +1,249 @@
...
+        log.debug("dataclass : %s", typename)
+
+        # Only update in broker its connected
+        for vss_signal_d in self.mapper.dds2vss_dict[topic_name].get("vsssignals"):
+            # One dds topic can hold data for multiple vss data points
+            element = vss_signal_d.get("element")
+            vss_signal = vss_signal_d.get("vsssignal")
+            log.debug("updating vss point %s", vss_signal)
+            log.debug("Picking data from element: %s", element)
+            # Update broker , TODO: Measure/evaluate making grpc calls in dds callback
```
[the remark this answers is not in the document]
lukasmittag: @AkhilTThomas do you plan to do this in the future?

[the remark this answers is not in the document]
AkhilTThomas: Yes, this is a wish.

dds2val/src/README.md
lukasmittag: I am no python import expert maybe some help from @romainletendart ;) ? If you have time :D Else maybe some thoughts from @AkhilTThomas ?

dds2val/src/ddsfeeder.py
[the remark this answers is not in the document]
lukasmittag: @AkhilTThomas could you then open an issue and describe this a little bit more?

dds2val/src/make_executable.sh
```diff
@@ -0,0 +1,3 @@
+if [[ "$TARGETARCH" == "arm64" ]]; then 
+  pyinstaller --clean --paths=. --add-binary="/opt/venv/lib/python3.9/site-packages/cyclonedds.libs/libddsc-62ba48e6.so.0.11.0:cyclonedds.libs/" --add-data="mapping.yml:." -F -s ddsfeeder.py; else pyinstaller --clean --paths=. --add-data="mapping.yml:." -F -s ddsfeeder.py
```
erikbosch, line 2: Risky as we do not specify which libddsc version we will have?

In [dbcfeeder](https://github.com/eclipse/kuksa.val.feeders/blob/main/dbc2val/Dockerfile#L37) I used a more generic solution to collect data. Would possibly `--collect-binaries cyclonedds` work?

dds2val/src/databroker.py
lukasmittag: Will be tackled in another PR

dds2val/src/README.md
erikbosch: Did you find a solution that satisfies both tests and docker? I do not know if restructuring like in dbcfeeder so that only the "main file" is in the top repo would help.

dds2val/src/make_executable.sh
lukasmittag: Problem here is it is collected by pyinstaller with this but DDS searches in a folder for it. But this is not ported by pyinstaller. Will ivestigate further.

dds2val/src/README.md
lukasmittag: Now it is working for me if I am in the dds2val directory and execute tests and ddsprovider.py or docker build from there

dds2val/OverallSequence.puml
```diff
@@ -0,0 +1,23 @@
...
+participant ddspublisher2
+end box
+ddspublisher1 --> DDS_network : publish dds message
+ddspublisher2 --> DDS_network : publish dds message
+alt on data reception
+ddsprovider --> databroker : update_datapoint
+end note
+end
+@enduml
+# was converted to Mermaid diagram, because Github can dynamically load mermaid diagrams in README
```
erikbosch, line 23: But then we can remove this file right, as content is directly included in README?

dds2val/ddsprovider.py
```diff
@@ -0,0 +1,53 @@
...
+    log.info("Starting ddsprovider...")
+
+    if os.environ.get("VEHICLEDATABROKER_DAPR_APP_ID"):
+        grpc_metadata = (
+            ("dapr-app-id", os.environ.get("VEHICLEDATABROKER_DAPR_APP_ID")),
+        )
+    else:
+        grpc_metadata = None
+
+    databroker_address = "127.0.0.1:" + os.environ.get("DAPR_GRPC_PORT", "55555")
```
erikbosch, line 26: Also support VDB_ADDRESS? Compare with https://github.com/boschglobal/kuksa.val.feeders/blob/feature/DDS_feeder/dbc2val/dbcfeeder.py#L423?

dds2val/tests/conftest.py
```diff
@@ -0,0 +1,36 @@
...
+#
+# This program and the accompanying materials are made available under the
+# terms of the Apache License 2.0 which is available at
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# SPDX-License-Identifier: Apache-2.0
+#################################################################################
+
+import pytest
+from py.xml import html
```
erikbosch, line 15: Pylint does not like this line. Seems to work without
```

erik@debian3:~/kuksa.val.feeders/dds2val$ pylint --exit-zero src tests
************* Module tests.conftest
tests/conftest.py:1:0: C0114: Missing module docstring (missing-module-docstring)
tests/conftest.py:15:0: E0401: Unable to import 'py.xml' (import-error)
tests/conftest.py:15:0: E0611: No name 'xml' in module 'py' (no-name-in-module)
tests/conftest.py:24:36: W0613: Unused argument 'call' (unused-argument)
tests/conftest.py:15:0: W0611: Unused html imported from py.xml (unused-import)
```

erikbosch (approved):
Looks good, only some minors that we possibly could ignore.

Possibly we should also mention ddsprovider on top level README (https://github.com/boschglobal/kuksa.val.feeders/blob/feature/DDS_feeder/README.md)

I did start ddprovider natively as well as from docker and did test from dds2val directory and all succeeded.

lukasmittag:
> Looks good, only some minors that we possibly could ignore.
> 
> Possibly we should also mention ddsprovider on top level README (https://github.com/boschglobal/kuksa.val.feeders/blob/feature/DDS_feeder/README.md)
> 
> I did start ddprovider natively as well as from docker and did test from dds2val directory and all succeeded.

@SebastianSchildt should we merge or would you like to see a config file?

dds2val/src/ddsfeeder.py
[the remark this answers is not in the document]
AkhilTThomas: > @AkhilTThomas could you then open an issue and describe this a little bit more?

https://github.com/eclipse/kuksa.val.feeders/issues/71#issue-1641353795

dds2val/README.md
```diff
@@ -0,0 +1,74 @@
...
+2. `pip install -r requirements-kml.txt`
+3. `cd kml && python3 dds_kmlreplay.py directions.kml`
+
+## How to run
+
+Choose from local build or contanerization via docker.
+These steps are necessary:
+
+1. Run an instance of databroker aka: `docker run -it --rm --net=host ghcr.io/eclipse/kuksa.val/databroker:master`
+2. Start the KML replay with an active local python virtual environment: `pip install requirements/requirements-kml.txt && cd kml && python3 dds_kmlreplay.py directions.kml`
```
erikbosch, line 30: This is almost the same as on line 20-22. If we are to keep it here we need to add ` -r` for `pip install`

dds2val/ddsprovider.py
```diff
@@ -0,0 +1,53 @@
...
+import signal
+import asyncio
+from pathlib import Path
+from ddsproviderlib import helper
+
+log = logging.getLogger("ddsprovider")
+
+async def main():
+    """Perform the main function activities."""
+    logging.basicConfig(level=logging.DEBUG)
```
erikbosch, line 12: We seem to ignore `LOG_LEVEL` that is set in Dockerfile. Should we possibly use INFO as default level.

dds2val/README.md
```diff
@@ -0,0 +1,74 @@
...
+
+## Configure the DDS provider
+
+Configuration for the DDS provider is solved through setting environment variables. Have a look at the table below.
+
+| Environment variable          | default value | description                                                                                                                                      |
+| ----------------------------- | ------------- | ------------------------------------------------------------------------------------------------------------------------------------------------ |
+| VEHICLEDATABROKER_DAPR_APP_ID | None          | DAPR ID for Vehicle App to look for. For more information to Vehicle Apps visit [Velocitas](https://eclipse-velocitas.github.io/velocitas-docs/) |
+| VDB_ADDRESS                   | 127.0.0.1     | Address where to look for (Vehicle) Databroker                                                                                                   |
+| DAPR_GRPC_PORT                | 55555         | Port where to look for (Vehicle) Databroker. If [DAPR](https://dapr.io/) gets used port of DAPR Sidecar where to look                            |
```
erikbosch, line 41: We use this even if we do not use DAPR. We could possibly call it something else but that could possibly be part of another PR.

erikbosch (approved):
Some minor ones. Did a test run with Docker and it works as expected

dds2val/README.md
lukasmittag: I think we can handle it like in dbc feeder use DAPR_GRPC_PORT to overwrite port for vehicle databroker. So I will introduce another environment variable VDB_PORT

PR #72, opened 2023-03-27
Fix config documentation

erikbosch:
Some ideas/questions on how to handle configs for clients

### DIFF
--- a/dbc2val/Readme.md
+++ b/dbc2val/Readme.md
@@ -104,18 +104,26 @@ A smaller excerpt from the above sample, with less signals.
 
 ## Configuration
 
-| parameter                     | default value   | config file          | Env var                       | command line argument | description                                                                                             |
-|-------------------------------|-----------------|----------------------|-------------------------------|-----------------------|---------------------------------------------------------------------------------------------------------|
-| config                        | -               | -                    | -                             | `--config`            | Configuration file                                                                                      |
-| dbcfile                       | -               | [can].dbc            | DBC_FILE                      | `--dbcfile`           | DBC file used for parsing CAN traffic                                                                   |
-| dumpfile                      | -               | [can].candumpfile    | CANDUMP_FILE                  | `--dumpfile`          | Replay recorded CAN traffic from dumpfile                                                               |
-| canport                       | -               | [can].port           | CAN_PORT                      | `--canport`           | Read from this CAN interface                                                                            |
-| use-j1939                     | False           | [can].j1939          | USE_J1939                     | `--use-j1939`         | Use J1939                                                                                               |
-| use-socketcan                 | False           | -                    | -                             | `--use-socketcan`     | Use SocketCAN (overriding any use of --dumpfile)                                                        |
-| mapping                       | mapping/vss_3.1.1/vss_dbc.json    | [general].mapping    | MAPPING_FILE                  | `--mapping`           | Mapping file used to map CAN signals to databroker datapoints. Take a look on usage of the mapping file |
-| server-type                   | kuksa_databroker | [general].server_type | SERVER_TYPE                 | `--server-type`       | Which type of server the feeder should connect to (kuksa_val_server or kuksa_databroker |
-| DAPR_GRPC_PORT                | -               | -                    | DAPR_GRPC_PORT                | -                     | Override broker address & connect to DAPR sidecar @ 127.0.0.1:DAPR_GRPC_PORT                            |
-| VEHICLEDATABROKER_DAPR_APP_ID | -               | -                    | VEHICLEDATABROKER_DAPR_APP_ID | -                     | Add dapr-app-id metadata                                                                                |
+| parameter                     | default value    | config file                   | Env var      | command line argument | description                                                                                             |
+|-------------------------------|------------------|-------------------------------|--------------|-----------------------|---------------------------------------------------------------------------------------------------------|
+| config                        | -                | -                             | -            | `--config`           | Configuration file  |
+| dbcfile                       | -                | [can].dbc                     | DBC_FILE     | `--dbcfile`          | DBC file used for parsing CAN traffic |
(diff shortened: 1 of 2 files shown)


dbc2val/Readme.md
```diff
@@ -104,18 +104,26 @@ A smaller excerpt from the above sample, with less signals.
...
+| ip (kuksa_databroker)         | 127.0.0.1        | [kuksa_databroker].ip         | VDB_ADDRESS,<br /> DAPR_GRPC_PORT | -| Databroker IP address, See below for Env var usage |
+| port (kuksa_databroker)       | 55555            | [kuksa_databroker].port       | VDB_ADDRESS,<br /> DAPR_GRPC_PORT | -| Databroker port, See below for Env var usage |
+| protocol (kuksa_databroker)   | grpc             | [kuksa_databroker].protocol   | -            | -                     | Configuration currently ignored |
+| insecure (kuksa_databroker)   | True             | [kuksa_databroker].insecure   | -            | -                     | Configuration currently ignored |
+| ip (kuksa_val_server)         | localhost        | [kuksa_val_server].ip         | -            | -                     | KUKSA.val Server IP address |
+| port (kuksa_val_server)       | 8090             | [kuksa_val_server].port       | -            | -                     | KUKSA.val Server port |
+| protocol (kuksa_val_server)   | ws               | [kuksa_val_server].protocol   | -            | -                     | Configuration currently ignored |
```
erikbosch, line 124: Currently we ignore "protocol". If server type is "kuksa_val_server" we use ws. Do we actually see a need to be able to specify what protocol to use? 

I see two options:

- Either remove the "protocol" config and let it be implied based on if you use server or broker
- Or keep it but give an error if it does not specify a supported protocol

dbc2val/Readme.md
```diff
@@ -104,18 +104,26 @@ A smaller excerpt from the above sample, with less signals.
...
+| strict                        | False            | -                             | -            | `--lax-dbc-strict`  | Disable strict parsing of dbc file |
+| server-type                   | kuksa_val_server | [general].server_type         | SERVER_TYPE  | `--server-type`     | Which type of server the feeder should connect to (kuksa_val_server or kuksa_databroker |
+| ip (kuksa_databroker)         | 127.0.0.1        | [kuksa_databroker].ip         | VDB_ADDRESS,<br /> DAPR_GRPC_PORT | -| Databroker IP address, See below for Env var usage |
+| port (kuksa_databroker)       | 55555            | [kuksa_databroker].port       | VDB_ADDRESS,<br /> DAPR_GRPC_PORT | -| Databroker port, See below for Env var usage |
+| protocol (kuksa_databroker)   | grpc             | [kuksa_databroker].protocol   | -            | -                     | Configuration currently ignored |
+| insecure (kuksa_databroker)   | True             | [kuksa_databroker].insecure   | -            | -                     | Configuration currently ignored |
```
erikbosch, line 121: No when we are to add authorization for data broker I think that "insecure" is a bad name to control TLS authentication.

Would it possible be better to rename it to `tls_auth` or similar? And if we want certificates/keys to be explicitly named - what is it that we want to provide? Just a folder-name and expected keys/certs to have specific names, or individual. If we currently just (want to) support client auth of server (i.e. not mutual auth) - then do we need anything more than a single file for the root cert? Which potentially could be used to imply if "insecure" should be used

And for Token authorization - there the client need only a single token file. 

So an idea I have - remove "insecure" and instead add two configs (with empty default)

- `tls_root_ca` - Path to root CA, if not given or empty "insecure" mode will be used
- `client_token` - If not given or empty the client will not try to authorize against broker/server

dbc2val/Readme.md
```diff
@@ -124,6 +132,25 @@ Configuration options have the following priority (highest at top).
 3. configuration file
 4. default value
 
+### USE_J1939
+
+This environment variable indicates if J1939 mode shall be used, and shall be `1` to indicate that J1939 mode shall be used.
+All other values will result in that J1939 mode is not used.
+
+
+### VDB_ADDRESS
```
erikbosch, line 141: Names of env-variables to give address and port seems to be quite inconsistent. In [databroker](https://github.com/eclipse/kuksa.val/tree/master/kuksa_databroker#configuration) `KUKSA_DATA_BROKER_ADDR` and `KUKSA_DATA_BROKER_PORT` is used, which differs a bit from what is used here. Do we actually need to have specific env-variables for Databroker and Server, respectively.

I am thinking for clients if we could just use `KUKSA_ADDR` and `KUKSA_PORT` and support them for both server and broker. Then we could also remove `DAPR_GRPC_PORT`, you could use the generic `KUKSA_ADDR` and `KUKSA_PORT` for that use case as well.

dbc2val/Readme.md
SebastianSchildt: `client-token`(with dash for consistency), or actually maybe just `token`(there is noch such thing as a server token in the system anyway) seems useful

For TLS, instead of "`--insecure` maybe just `--no-tls`

Insecure mode should not be used when when no root cert is given, because the dafualt behavior should be, that hte TLS implementation refers to the system Root certs (as it will do "out of the box" anyway) 

So, even without `--tls-ca-cert` it should attempt tls connection unless `--no-tls`is given


Also, I think the plumbing for optionally giving custom root-ca-certs  needs to be done in kuksa-client. An alternative could be not implementing that, and instead say, people should configure the root-ca store of their system accordingly, but on the other that makes testings very unpractical.

dbc2val/Readme.md
SebastianSchildt: No real opinion here. Sounds reasonable. Not a big user of env vars, but they can come in handy in K8/3/0S deployments

dbc2val/Readme.md
erikbosch: @SebastianSchildt -digging a bit deeper into code it appears that these configs actually are passed "as is" to kuksa-client in https://github.com/eclipse/kuksa.val/blob/master/kuksa-client/kuksa_client/cli_backend/__init__.py#L25, so we cannot easily change name in the feeders. It also the same file that specifies that there shall be a config "protocol"

Issue #73, opened 2023-04-04
Error "No module named 'can.interfaces.socketcan'" when using dbc2val docker container v0.1.1


mikehaller:
I am using `ghcr.io/eclipse/kuksa.val.feeders/dbc2val:v0.1.1` and run it like this:

```
docker run --rm -it \
    -v `pwd`:/custom \
    --privileged \
    --network host \
    ghcr.io/eclipse/kuksa.val.feeders/dbc2val:v0.1.1 \
    ./dbcfeeder-exe \
    --use-socketcan \
    --canport vcan0 \
    --usecase databroker \
    --dbcfile /custom/custom.dbc \
    --mapping /custom/custom_vss_dbc.json
```

The resulting error message is:

```
2023-04-04 09:41:49,222 INFO dbc2vssmapper: Mapper: No minimal update delay defined for signal Vehicle, setting to 1000ms.
2023-04-04 09:41:49,222 INFO dbcreader: Reading DBC file /custom/custom.dbc
2023-04-04 09:41:50,749 INFO dbcreader: Collecting signals, generating CAN ID whitelist
2023-04-04 09:41:50,751 WARNING dbcreader: Signal Vehicle not found in DBC file
2023-04-04 09:41:50,751 INFO dbcreader: CAN ID whitelist=[]
Traceback (most recent call last):
  File "dbcfeeder.py", line 530, in <module>
  File "dbcfeeder.py", line 446, in main
  File "dbcfeeder.py", line 140, in start
  File "dbcreader.py", line 56, in start_listening
  File "can/interface.py", line 104, in __new__
  File "can/interface.py", line 42, in _get_class_for_interface
can.exceptions.CanInterfaceNotImplementedError: Cannot import module can.interfaces.socketcan for CAN interface 'socketcan': No module named 'can.interfaces.socketcan'
[7] Failed to execute script 'dbcfeeder' due to unhandled exception!
```

Expected:
- dbcfeeder container image can be used with SocketCAN, e.g. `vcan0` on the host

mikehaller:
After modifying the `Dockerfile` to let pyinstaller know about missing modules using `--hidden-import`, it seems to work.

Changes to the dockerfile:
```
RUN pyinstaller --collect-data kuksa_certificates --hidden-import can.interfaces.socketcan --clean -F -s dbcfeeder.py
```

Tested with vcan0 on the host, using cantools python library to send a CAN Frame.
[...]

PR #75, opened 2023-04-05
Fix missing socketcan in dbc2val

mikehaller:
pyinstaller seems to be unable to detect the CAN library at build time. At runtime, the dbc2val cannot use the "--use-socket-can" feature as the library is missing. Informing pyinstaller about the hidden import let's it package the CAN library properly.


mikehaller:
Fixes #73

erikbosch (approved):
Tested building a docker container and running with canplayer, works.

PR #76, opened 2023-04-05
Update DBC provider action file to kuksa.val standard

SebastianSchildt:
This updates DBC feeder CI, to use the same patterns and version as databroker,

Will push images for testing to ttl.sh, and should push correctly tagged images to ghcr.io upon workflow dispatch with a tag

dbc tests and pylint still in/not modified

### DIFF
--- a/.github/workflows/kuksa_dbc_feeder.yml
+++ b/.github/workflows/kuksa_dbc_feeder.yml
@@ -1,5 +1,5 @@
 # /********************************************************************************
-# * Copyright (c) 2022 Contributors to the Eclipse Foundation
+# * Copyright (c) 2022,2023 Contributors to the Eclipse Foundation
 # *
 # * See the NOTICE file(s) distributed with this work for additional
 # * information regarding copyright ownership.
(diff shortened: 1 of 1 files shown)


.github/workflows/kuksa_dbc_feeder.yml
```diff
@@ -27,65 +27,83 @@ jobs:
...
-          push: false
-          outputs: |
-            type=oci,dest=./dbc2val.tar
-          context: ./dbc2val
-          file: ./dbc2val/Dockerfile
-          platforms: linux/amd64, linux/arm64
-          tags: ${{ github.sha }}
-          labels: |
-            org.opencontainers.image.source=https://github.com/${{steps.repository-name-adjusted.outputs.lowercase}}
+    - name: Build ephemereal DBC provider container and push to ttl.sh
```
erikbosch, line 78: Should it be "ephemeral", not "ephemereal"?

SebastianSchildt: Should.
Done.

PR #77, opened 2023-04-06
adds CSV provider

eriksven:
This PR adds a provider which reads VSS signals from a given CSV-file and replays this data and the changes over time to an instance of the kuksa.val databroker.

### DIFF
--- a/csv_provider/README.md
+++ b/csv_provider/README.md
@@ -1,5 +1,6 @@
 # VSS Provider
-The aim of this script to provide and replay VSS data into a `kuksa.val` databroker. It is possible to define a sequence of VSS signals including timing delays in a CSV-file and then to apply this sequence towards a `kuksa.val` databroker with the provider.
+The aim of this script is to provide and replay VSS data into a `kuksa.val` databroker.
+Therefore, it takes a CSV-file, containting pre-defined sequences of VSS signals including timing delays, and provides it to the 'kuksa.val' databroker.
 
 ## Usage
 The provider requires an installation of Python in version 3 and can be executed with the following commands:
--- a/csv_provider/README.md
+++ b/csv_provider/README.md
@@ -1,5 +1,17 @@
 # VSS Provider
-The aim of this script to provide and replay VSS data into a `kuksa.val` databroker. It is possible to define a sequence of VSS signals including timing delays in a CSV-file and then to apply this sequence towards a `kuksa.val` databroker with the provider. 
+The aim of this script to provide and replay VSS data into a `kuksa.val` databroker. It is possible to define a sequence of VSS signals including timing delays in a CSV-file and then to apply this sequence towards a `kuksa.val` databroker with the provider.
+
+## Usage
+The provider requires an installation of Python in version 3 and can be executed with the following commands:
+
+```
+pip install -r requirements.txt
+python provider.py
+```
+
+This assumes a running `kuksa.val` databroker at `127.0.0.1:55555` and a file named `signals.csv` containing the signals to apply. See the section [Arguments](#arguments) for more details on possible arguments and default values.
+
+The provider uses the [kuksa_client]() Python implementation which you need to install on your system, e.g., by applying the [requirement.txt](requirements.txt) with pip.
 
 ## Arguments
 You can start the provider with the following arguments on a command line:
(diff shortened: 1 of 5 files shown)


csv_provider/README.md
```diff
@@ -0,0 +1,31 @@
...
+## Arguments
+You can start the provider with the following arguments on a command line:
+
+| short argument | long argument | description | default value |
+|---- | ---- | ----- | ----|
+|-f| --file | This indicates the CSV-file containing the signals to update in the `kuksa.val` databroker. | signals.csv |
+| -a | --address | This indicates the address of `kuksa.val` databroker to connect to. | 127.0.0.1 |
+| -p | --port | This indicates the port of the `kuksa.val` databroker to connect to. | 55555 |
+| -i | --infinite | If the flag is set, the provider loops over the file until stopped, otherwise the file gets processed once. | not present/False
+| -l | --log | This sets the logging level. Possible values are: DEBUG, INFO, DEBUG, WARNING, ERROR, CRITICAL | WARNING
```
erikbosch, line 13: The list of possible values is a bit difficult to understand. Is it deliberate that DEBUG and WARNING occurs twice?

csv_provider/README.md
```diff
@@ -0,0 +1,31 @@
...
+The delimiter for the CSV-file is the ',' sign and there is no header in the file, meaning that the first line is interpreted as application data. 
+
+Each line in the csv file consists of the following elements:
+
+| index | description | example |
+| -- | -----------| --|
+|0 | indicates whether to update the current value (`current`) or the target value (`target`) for the signal. | current |
+| 1 | the name of the signal to update | Vehicle.Speed
+| 2 | the new value | 48 |
+| 3 | Indicates the time in seconds which the provider shall wait after processing this signal. This way one can emulate the change of signals over time. | 0 |
```
erikbosch, line 31: I think it would be good to include an example on how you use the provider with the example file. The example could also include how you call pip to install dependencies

csv_provider/requirements.txt
```diff
@@ -0,0 +1 @@
+kuksa-client==0.3.0
```
SebastianSchildt, line 1: Maybe better  say "minimum 0.3" (>=), or at least 0.3.x (~=)

csv_provider/signals.csv
SebastianSchildt, line 1: As this is an examaple, I would also add an header giving a short name/description about the contents. This is a convention often used in CSV, that the first line is basically the header/legend for what is in hte columns.  The python csv reader has some heuristics, that should detect it just fine.

SebastianSchildt:
Not really something, that needs to be added here, just an idea: a helper script creating these .csv from a "life" system is probably easy to do: Give it a list of VSS paths, and it can depending on metadata, subscribe current value for sensors and current&target value for actuators while writing a CSV file. (Don't forget to flush on Ctrl-C :) ).  But just an idea for the future

csv_provider/README.md
eriksven: No, having DEBUG twice was simply an error. 

To align with the dbcfeeder, I also set the default logging level to DEFAULT and downgraded some of the logging.

csv_provider/requirements.txt
eriksven: i opted for 0.3.x

csv_provider/signals.csv
eriksven: I added a header now. Initially, I opted against adding a header to make it easier to generate the CSV-file with other tooling. But you are right that having the header improves the usability of the files. In addition, having the header allows to use Dicts instead of an array to represent each row making the access to the values a bit easier to understand in the code.

csv_provider/provider.py
```diff
@@ -0,0 +1,69 @@
...
+                    logging.info("Update target value of %s to %s", row[1], row[2])
+                else:
+                    updates = []
+                
+                try:
+                    await client.set(updates=updates)
+                except Exception as ex:
+                    logging.error("Error while updating %s\n%s", row[1], ex)
+                
+                await asyncio.sleep(int(row[3]))
```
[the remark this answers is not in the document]
eriksven: Good point, I changed the cast to float and also added an error message if the delay value cannot be processed.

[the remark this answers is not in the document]
eriksven: Good point, I changed the cast to float and also added an error message if the delay value cannot be processed.

csv_provider/README.md
eriksven: After reconsidering the logging levels, I decided to not downgrade to DEBUG when sending a signal because in DEBUG the kuksa_client library also adds an extensive log message. So keeping the level of INFO in this case allows to decide between the extensive log and the simple information that the line in the CSV-file has been processed.

eriksven:
I updated the branch for the CSV-provider, @erikbosch and @SebastianSchildt could you please have another look. 

The idea of having a script to generate the CSV-files from a life system is good. To me this is beyond the scope of this PR but I will have look into it for a further PR.

csv_provider/README.md
```diff
@@ -0,0 +1,43 @@
+# VSS Provider
+The aim of this script to provide and replay VSS data into a `kuksa.val` databroker. It is possible to define a sequence of VSS signals including timing delays in a CSV-file and then to apply this sequence towards a `kuksa.val` databroker with the provider.
```
lukasmittag, line 2:
```suggestion
The aim of this script is to provide and replay VSS data into an instance of `kuksa.val` databroker. Therefore it takes a CSV file, containing pre defined sequences, consisting of VSS signals including timing delays, and provides it to the 'kuksa.val' databroker
```

lukasmittag (approved):
Just some minor comment regarding documentation. Decide for yourself if you want to rephrase or not.

csv_provider/README.md
eriksven: Thanks for the suggestion, I adapted the paragraph based on it.

PR #78, opened 2023-04-06
Do not load CAN log into memory when replaying

sophokles73:
Currently, a CAN dump file is being loaded into memory before starting to replay its frames to the databroker.
This PR changes the behavior to read the CAN frames from the dump file one-by-one instead of pre-loading the whole file into memory.

### DIFF
--- a/dbc2val/dbcfeederlib/canplayer.py
+++ b/dbc2val/dbcfeederlib/canplayer.py
@@ -13,8 +13,6 @@
 
 import logging
 import threading
-import time
-
 import can
 
 # fixes issues with pytinstaller not detecting can.interfaces.virtual usage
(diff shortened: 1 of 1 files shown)


dbc2val/dbcfeederlib/canplayer.py
```diff
@@ -40,73 +38,42 @@ class CANplayer:
...
+                log.debug("Failed to send message via CAN bus")
 
-    def getNextMessage(self):
-        val = self.messages[self.indexout]
-        self.indexout += 1
-        # in case of end of array set the index pointer to start again
-        if self.indexout >= self.index:
-            self.indexout = 1
-        return val
+        log.debug("Replayed all messages from CAN log file")
```
erikbosch, line 61: I think this could be a log.info, can be quite useful and shall only be given once, right?

sophokles73: this is logged each time the log has been re-played fully.

dbc2val/dbcfeederlib/canplayer.py
```diff
@@ -40,73 +38,42 @@ class CANplayer:
...
-        txThread = threading.Thread(target=self.txWorker)
-        txThread.start()
+            if not self.run:
+                break
+            try:
+                self.bus.send(msg)
+                log.debug(f"Message sent on {self.bus.channel_info}")
+                log.debug(f"Message: {msg}")
+            except can.CanError:
+                log.debug("Failed to send message via CAN bus")
```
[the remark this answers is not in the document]
sophokles73: no particular reason, I tend to keep log statements at a low level to reduce verbosity. However, in this case it probably makes sense to raise the level accordingly. We expect this to happen very seldomly only, I guess.

sophokles73:
@erikbosch I have adjusted the log levels

dbc2val/dbcfeederlib/canplayer.py
erikbosch: Ah, then it can actually come quite often if we run the log repeatedely

erikbosch (approved):
Did a sanity test (native together with databroker) and it worked without problem, ok for me

SebastianSchildt (approved):
Nice. I wanted this for a long time, never came around to it. Now even the the huge logs starts instantaneously. And less lines. Seems I can have my 🍰  and eat it too

PR #79, opened 2023-04-06
Some improvements to dbc2val README

sophokles73:
Improved readability of the config parameter table and added missing parameter for setting Databroker IP:PORT.
Also aligned usage of the *USE_J1939* environment variable with how the *--use-j1939* switch works.


erikbosch (approved):
Looks OK to me

PR #82, opened 2023-04-12
Also built and push a DBC provider container on merges to main

SebastianSchildt:
So we also have a master/main container similar to databroker etc.


erikbosch:
I guess we either need to merge https://github.com/eclipse/kuksa.val.feeders/pull/83 or specify `pyinstaller==5.9.0` in Dockerfile to make CI pass. When build succeeds I am happy to merge this one

PR #83, opened 2023-04-12
Remove staticx

SebastianSchildt:
I think "size-wise" we can get rid of staticx, and this is one quick shot at it.

Advantages
 - no more weird patchelf issues or incompatibilities with pyintaller
 - one less moving part

Disadvantages
 - To get a small image needed to switch to Alpine (as "from scratch" is not possible after pyinstaller). This leads to longer build times due to GRPC needing to be built for MUSL

The current size on arm64

```
 % docker images | grep dbc                                         
dbc                                                                              latest                 a0f3a5b42edb   3 minutes ago    39.4MB
```

I did test, it can start and play logfile to dataroker, I had no way to
 - test amd64 version
 - test whether access ot socketcan (vcan) still works


dbc2val/Dockerfile
```diff
@@ -31,35 +33,26 @@ ENV PATH="/opt/venv/bin:$PATH"
 RUN /opt/venv/bin/python3 -m pip install --upgrade pip \
     && pip3 install --no-cache-dir -r requirements.txt
 
-RUN pip3 install wheel scons && pip3 install pyinstaller patchelf==0.17.0.0 staticx
+RUN pip3 install wheel scons && pip3 install pyinstaller
```
erikbosch, line 36: Based on our recent problems with pyinstaller (well actually with staticx it seems) I am thinking if we in the long run should handle the dependencies here similar to our requirements* files, i.e. directly or indirectly specify exact version so we can always "rebuild" the same docker container. Maybe not for main/master, but maybe for releases.

erikbosch (approved):
I did the following:

- Built the container for amd64
- Run it on my Linux VM towards Databroker, with and without CANplayer running separately.

Both worked, so functionally I am in favor of this PR

dbc2val/Dockerfile
```diff
@@ -13,14 +13,16 @@
...
-FROM --platform=$TARGETPLATFORM python:3.9-slim-bullseye as builder
+FROM --platform=$TARGETPLATFORM python:3.10-alpine as builder
 
 ARG TARGETPLATFORM
 ARG BUILDPLATFORM
 
 RUN echo "-- Running on $BUILDPLATFORM, building for $TARGETPLATFORM"
 
-RUN apt-get update -qqy && apt-get upgrade -qqy && apt-get install -qqy binutils g++
+#‚RUN apt-get update -qqy && apt-get upgrade -qqy && apt-get install -qqy binutils g++
```
erikbosch, line 23: Still draft, but some cleanup required before merge

dbc2val/Dockerfile
SebastianSchildt: Should we add it to dev-dependencies? Or create a seperate "requirements-package.in" or something, generate it and pip-install it as well?

Issue #86, opened 2023-04-14
[dbc2val] Container image build fails


sophokles73:
Trying to build the container image using the Dockerfile yields the following error:
```sh
=> ERROR [builder 10/14] RUN staticx --debug dbcfeeder dbcfeeder-exe                                                                                               1.3s 
------                                                                                                                                                                   
 > [builder 10/14] RUN staticx --debug dbcfeeder dbcfeeder-exe:                                                                                                          
#0 0.792 INFO:root:Running StaticX version 0.13.8                                                                                                                        
#0 0.792 INFO:root:Libraries:                                                                                                                                            
#0 0.793 INFO:root:  elftools: 0.29                                                                                                                                      
#0 0.793 DEBUG:root:External tools:                                                                                                                                      
#0 0.797 INFO:root:  ldd: /usr/bin/ldd: ldd (Debian GLIBC 2.31-13+deb11u5) 2.31
#0 0.801 INFO:root:  objcopy: /usr/bin/objcopy: GNU objcopy (GNU Binutils for Debian) 2.35.2
#0 0.803 INFO:root:  strip: /usr/bin/strip: GNU strip (GNU Binutils for Debian) 2.35.2
#0 0.805 INFO:root:  patchelf: /opt/venv/bin/patchelf: patchelf 0.17.0
#0 0.805 DEBUG:root:Arguments:
#0 0.806 DEBUG:root:  prog:      'dbcfeeder'
#0 0.806 DEBUG:root:  output:    'dbcfeeder-exe'
#0 0.806 DEBUG:root:  libs:      None
#0 0.806 DEBUG:root:  strip:     False
#0 0.806 DEBUG:root:  compress:  True
#0 0.806 DEBUG:root:  debug:     True
#0 0.807 INFO:root:Using XZ BCJ filter FILTER_X86
#0 0.828 DEBUG:root:Bootloader: bootloader version 0.13.8 compiled Aug  7 2022 at 04:42:01 by musl-gcc version 10.2.1 20210110
#0 0.830 INFO:root:Program interpreter: /lib64/ld-linux-x86-64.so.2
#0 0.864 DEBUG:root:Running ['patchelf', '--remove-rpath', '/tmp/staticx-prog-nd9zcje8']
[...]

PR #87, opened 2023-04-14
[dbc2val] Pin pyinstaller version used for container image build

sophokles73:
This is for #86

The pyinstaller version has been pinned to 5.9.0 to prevent usage of the recently released 5.10.0 version which contains an internal API change that causes staticx to break.

See https://github.com/JonathonReinhart/staticx/issues/235

Also changed the Dockerfile to install the standard patchelf OS package instead of using pip install because this is the approach recommended by the staticx installation guide.

### DIFF
--- a/dbc2val/Dockerfile
+++ b/dbc2val/Dockerfile
@@ -1,5 +1,5 @@
 # /********************************************************************************
-# * Copyright (c) 2022 Contributors to the Eclipse Foundation
+# * Copyright (c) 2022,2023 Contributors to the Eclipse Foundation
 # *
 # * See the NOTICE file(s) distributed with this work for additional
 # * information regarding copyright ownership.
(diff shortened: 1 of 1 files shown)


erikbosch:
We need to improve why we do things. The patchelf pinning was recently added by https://github.com/eclipse/kuksa.val.feeders/pull/50 but we do not really provide more information on why it was needed. @lukasmittag  - can you check if the reason for a pinned version is still relevant.

@sophokles73 - @SebastianSchildt has another solution addressing the same problem in https://github.com/eclipse/kuksa.val.feeders/pull/83. Could you possibly check if that solution fits your needs as well? I guess that either that one or this one shall be merged.

sophokles73:
> @SebastianSchildt has another solution addressing the same problem in https://github.com/eclipse/kuksa.val.feeders/pull/83. Could you possibly check if that solution fits your needs as well? I guess that either that one or this one shall be merged.

I had indeed seen it and its approach should definitely work. This PR was intended as a quick fix until #83 is being merged ...

sophokles73:
fixed via https://github.com/eclipse/kuksa.val.feeders/commit/2feefd14701811020ccb59ec2efe9a1a6ccd0b31

PR #90, opened 2023-04-18
Update DDS provider to use authorization from kuksa-client lib

### LINKED ISSUE #80: Support authorization in ddsfeeder
As we now have added a ddsfeeder we need to add authorization support for databroker there as well

Can possibly resue parts of: https://github.com/eclipse/kuksa.val.feeders/pull/74
Relies on that https://github.com/eclipse/kuksa.val/pull/517 is merged and a new kuksa-client python package released

lukasmittag:


erikbosch:
Seems it fails on kuksa-client version. IF we want to test pass we need to specify version > 0.3 and use pip with `--pre`

.github/workflows/kuksa_dds_feeder.yml
erikbosch, line 1: There are some trailing blanks on line 23, that is why the pre-commit check fails. (It checks all files modified)

dds2val/README.md
```diff
@@ -41,6 +41,7 @@ Configuration for the DDS provider is solved through setting environment variabl
 | DAPR_GRPC_PORT                | None          | If [DAPR](https://dapr.io/) gets used port of DAPR Sidecar. Overwrites VDB_PORT variable                                                         |
 | MAPPING_FILE                  | mapping.yml   | Place of mapping file from DDS to VSS                                                                                                            |
 | VDB_PORT                      | 55555         | On which port the (vehicle) databroker is expected. If you want to use DAPR use DAPR_GRPC_PORT.                                                  |
+| TOKEN                         | None          | JWT token which will get used to authorize to databroker                                                                                         |
```
erikbosch, line 44: So here it must be a token, giving a file path to a token does not work, or? If so - would it be valuable with a "real example" on how to use/start dds-feeder with authentication, i.e. giving token as env-variable?

lukasmittag:
Yes, example would be good I guess. And yes it supports only giving a token not a tokenfile. It could be changed to accept both, if we want this.

erikbosch:
> Yes, example would be good I guess. And yes it supports only giving a token not a tokenfile. It could be changed to accept both, if we want this.

As long as it is well defined what we support I think it is fine. And an example, based on using one of the "default tokens" would be good, if nothing else so that we know ourselves how to test it. I assume you on Linux can use some fancy one line command to set the environment variable with the content of the file.

erikbosch (approved):

PR #96, opened 2023-05-05
Adapting to VSS 4.0

erikbosch:
Adapting to VSS 4.0 based on release candidate.


dds2val/ddsproviderlib/idls/generate_py_dataclass.sh
```diff
@@ -2,17 +2,17 @@
...
+# moving files to tmp as the py files generated as on the same folder
 # as the package name and would get deleted
 
 # git clone Vechicle signal specidification and vss-tools
 cd /tmp
 rm -rf /tmp/vehicle_signal_specification
 rm -rf /tmp/vss-tools
-git clone https://github.com/COVESA/vehicle_signal_specification.git -b v3.1.1
-git clone https://github.com/COVESA/vss-tools.git -b v3.1
+git clone https://github.com/COVESA/vehicle_signal_specification.git -b v4.0rc0
```
erikbosch, line 12: NOTE: This needs to be changed to real tag later

We also have possibly found a bug in ddsidl exporter, build will likely fail

dds2val/ddsproviderlib/idls/generate_py_dataclass.sh
```diff
@@ -2,17 +2,17 @@
...
 # as the package name and would get deleted
 
-# git clone Vechicle signal specidification and vss-tools
+# git clone Vehicle signal specification and vss-tools
 cd /tmp
 rm -rf /tmp/vehicle_signal_specification
 rm -rf /tmp/vss-tools
-git clone https://github.com/COVESA/vehicle_signal_specification.git -b v3.1.1
-git clone https://github.com/COVESA/vss-tools.git -b v3.1
+git clone https://github.com/COVESA/vehicle_signal_specification.git -b v4.0rc1
```
erikbosch, line 12: This is to be changed before final release

erikbosch:
Now things seems to work, but some refactoring in ddsprovider required to support use-cases where we cannot use "latest master" of databroker. I think this one is ready for an internal review. I have verified ddsprovider and dbcfeeder.

dbc2val/test/test_example_mapping/test_example_mapping.py
```diff
@@ -76,55 +75,56 @@ def test_parking_lock():
...
+
     # For non successful mapping old value shall be kept and None returned
     result = vss_mapping.transform_value("Something undefined")
-    assert result == None
+    assert result is None
+
 
 def test_vehicle_speed():
     dbc_mapping = mapper["DI_uiSpeed"]
     assert len(dbc_mapping) == 2
```
lukasmittag, line 101: do we want is here instead of == like you changed at the lines above (True or False)?

dbc2val/test/test_example_mapping/test_example_mapping.py
```diff
@@ -76,55 +75,56 @@ def test_parking_lock():
...
-    
+
     if dbc_mapping[0].vss_name == "Vehicle.Speed":
         speed_index = 0
     else:
         speed_index = 1
     vss_mapping = dbc_mapping[speed_index]
     assert vss_mapping.vss_name == "Vehicle.Speed"
```
lukasmittag, line 108: same here

dbc2val/test/test_example_mapping/test_example_mapping.py
```diff
@@ -76,55 +75,56 @@ def test_parking_lock():
...
+    assert vss_mapping.transform is None
+
     # Now test some transforms
     result = vss_mapping.transform_value(25.3)
     assert result == 25.3
 
+
 def test_vehicle_obd_speed():
     dbc_mapping = mapper["DI_uiSpeed"]
     assert len(dbc_mapping) == 2
```
lukasmittag, line 118: and here

dbc2val/test/test_example_mapping/test_example_mapping.py
```diff
@@ -76,55 +75,56 @@ def test_parking_lock():
...
     vss_mapping = dbc_mapping[speed_index]
     assert vss_mapping.vss_name == "Vehicle.OBD.Speed"
-    assert vss_mapping.transform == None
-    
+    assert vss_mapping.transform is None
+
     # Now test some transforms
     result = vss_mapping.transform_value(25.3)
     assert result == 25.3
-
```
lukasmittag, line 130: no new line at end of file intended?

dbc2val/test/test_example_mapping/test_example_mapping.py
erikbosch: `is` compares if two objects are the same. `==` if two objects are equal. It seems `is` is preferred when comparing with `None`, and at least some linters wants it for comparison with `True` and `False` as well. But for integers it only seems to work for smaller values where type object is cached. So my understanding is that `==` is recommended for integer comparison

https://realpython.com/courses/python-is-identity-vs-equality/
https://stackoverflow.com/questions/132988/is-there-a-difference-between-and-is

dbc2val/test/test_example_mapping/test_example_mapping.py
erikbosch: When looking in an editor there seems to be a newline. What previously existed was a number of blanks on last line and no end-of-line after the blanks. Those blanks have been removed, and this seems to be how github presents it. So the resulting file seems to me to be corrected.

erikbosch:
@lukasmittag - could you check my answers above - if there is any remaining fixes to do?

dbc2val/test/test_example_mapping/test_example_mapping.py
lukasmittag: I was just wondering if with a string compare it would be nicer to have is. Would be more readable. @erikbosch if you do not want to change/disagree we can merge.

erikbosch: `==` seems to be preferred for strings.

See https://stackoverflow.com/questions/67489572/should-i-use-for-string-comparison

If using `is` flake8 complains:
```

erik@debian3:~/kuksa.val.feeders$ git commit --amend
Trim Trailing Whitespace.................................................Passed
Fix End of Files.........................................................Passed
Check Yaml...........................................(no files to check)Skipped
Check for added large files..............................................Passed
flake8...................................................................Failed
- hook id: flake8
- exit code: 1

dbc2val/test/test_example_mapping/test_example_mapping.py:120:8: F632 use ==/!= to compare constant literals (str, bytes, int, float, tuple)
```

PR #97, opened 2023-05-08
For conenction to websocket authorize method needs to be called

lukasmittag:
(python library authorizes on startup to grpc interface)


erikbosch (approved):
Looks good, more or less the same as I realized as investigating VSS 4.0 support in https://github.com/eclipse/kuksa.val.feeders/pull/96/files#diff-cc82792cd24a1dcb2b9e3ff1cd0aa304411e6ae2c57cd103e5570e8d56e4f470

PR #98, opened 2023-05-10
Updating actions to avoid Node.js 12

erikbosch:
Origin is a github warning:
Node.js 12 actions are deprecated.
Please update the following actions to use Node.js 16: docker/setup-buildx-action@v1, docker/login-action@v1, ASzc/change-string-case-action@v2, docker/setup-qemu-action@v1, docker/build-push-action@v2.
For more information see:
https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/.

I have updated all actions to latest and greatest. No testing performed.


erikbosch:
@int0x27 - there seems to be some someip-related problem, they do not seem to be related to my changes but rather "regressions" when using a new vsomeip version and a newer conan-version. Known problems?

.github/workflows/kuksa_dds_feeder.yml
```diff
@@ -36,27 +36,27 @@ jobs:
...
+      uses: docker/login-action@v2
       with:
         registry: ghcr.io
         username: ${{ github.repository_owner }}
         password: ${{ secrets.GITHUB_TOKEN }}
 
     - id: repository-name-adjusted
       name: Make repository name in lower case for docker upload.
-      uses: ASzc/change-string-case-action@v2
+      uses: ASzc/change-string-case-action@v5
```
erikbosch, line 50: By the way - for which repository name do we need to change case? This action seems to be used in some *.yml files but not all. Needed?

erikbosch:
Needs to wait for #101 and then be rebased

.github/workflows/kuksa_someip_feeder_release.yml
```diff
@@ -58,7 +58,7 @@ jobs:
     steps:
       - id: repository-name-adjusted
         name: Prepare repository name in lower case for docker upload. This supports repository names in mixed case
-        uses: ASzc/change-string-case-action@v2
```
int0x27, line 61: this rule is not was replaced by someip2val_build.yml. not used

erikbosch:
Blocked by #96. Current error in dds test depends on that latest master of broker now uses 4.0, but dds still use 3.1, and there is a slight difference in signals used.

(Which brings up the topic of how to handle dependencies between kuksa.val and kuksa.val.feeders. If we consider feeder to be downstream then we should at least for releases refer to released/tagged versions of kuksa.val)

erikbosch:
Block removed, ready to be merged

int0x27 (approved):
Looks good

PR #99, opened 2023-05-10
Adding license check

erikbosch:
Reusing license check from KUKSA.val

For a real example with changed files see https://github.com/eclipse/kuksa.val.feeders/pull/96


lukasmittag (approved):
Looks good, will merge

PR #100, opened 2023-05-11
Align SPDX header for Python files

erikbosch:
This is to make sure all our Python files fulfill our current SPDX check

This PR do:

- Adds SPDX to all python files that did not had it before
- Add our generic copyright to those files (without SPDX) that did not had it before. In some cases this means that explicit Bosch copyright is removed
- Adds python3 shebang for all executables. 
- Also making all "top level" python files executable.

### DIFF
--- a/csv_provider/provider.py
+++ b/csv_provider/provider.py
@@ -1,3 +1,18 @@
+#! /usr/bin/env python3
+
+########################################################################
+# Copyright (c) 2023 Contributors to the Eclipse Foundation
+#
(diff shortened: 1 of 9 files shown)


SebastianSchildt (approved):
Not a fan of shebang with space. But "the Internet" says it is perfectly fine https://unix.stackexchange.com/questions/276751/is-space-allowed-between-and-bin-bash-in-shebang 


lgtm 🦛

PR #101, opened 2023-05-12
Added Authorization support for someip2val

### LINKED ISSUE #68: Support authorization in SOME/IP feeder
What needs to be done:
* Introduce authorization in https://github.com/eclipse/kuksa.val.feeders/blob/main/someip2val/src/someip_feeder/someip_kuksa_adapter.cc
* Think of a good default? token and add argument, config to specify token/tokenfile

int0x27:
resolves #68 

- Added target actuator subscriber using kuksa common GRPC API.
- JWT token authorization is implemented in `CollectorClient`.
- `cert/` contains someip2val specific JWT token and readme how to get required keys from KUKSA.val.
- Added `--help` argument for `someip_feeder` showing options and used environment variables.
- Added common `simple_log` for broker_feeder classes.
- Added env `WIPER_STATUS` to mute `###|WiperEvent|` status lines (or make them print on the same line).
- Fixed vsomeip tag to `3.1.20.3` as `master` is no longer compatible.


int0x27:
Need to sync proto files with kuksa.val:

https://github.com/boschglobal/kuksa.val.feeders/blob/db45fbfdfb0b1ac03be2094eb96ed9a2e13dc03a/someip2val/proto/kuksa_data_broker/sdv/databroker/v1/collector.proto#L61

.github/workflows/someip2val_build.yml
```diff
@@ -56,7 +52,7 @@ jobs:
               cmake g++ build-essential \
               g++-aarch64-linux-gnu binutils-aarch64-linux-gnu \
               jq python3 python3-pip python3-setuptools &&
-          pip3 install -U conan
+          pip3 install conan==1.55.0
```
erikbosch, line 55: Would it be an idea to add a comment on why we want to pin conan.
I see a risk that sooner or later we will get a vulnerability warning and then it is good to know why we pinned it.

Is it only because newer conan does not support "info"?

int0x27: 2.0 broke arm64 cross compiling and also introduce a lot of new recipes, not sure if the state is good enough, but I also fixed the revisions of some important dependencies that only work with specific conan version. It's better to get a security warning than to find some day we are not building at all

someip2val/src/lib/broker_feeder/collector_client.cc
```diff
@@ -11,55 +11,81 @@
...
     , connected_(false)
 {
+    // set log level from env "KUKSA_DEBUG", defaults to info
+    LOGGER_SET_LEVEL_ENV("KUKSA_DEBUG", LEVEL_INF);
+
     changeToDaprPortIfSet(broker_addr);
     metadata_ = getGrpcMetadata();
     channel_ = grpc::CreateChannel(broker_addr, grpc::InsecureChannelCredentials());
     stub_ = sdv::databroker::v1::Collector::NewStub(channel_);
+    kuksa_stub_ = kuksa::val::v1::VAL::NewStub(channel_);
```
erikbosch, line 47: How is the design? Do we use both APIs (sdv::databroker::v1 and kuksa::val::v1) in parallel? What is the benefit of that, or why is it needed?

someip2val/proto/kuksa_data_broker/kuksa/val/v1/types.proto
```diff
@@ -0,0 +1,286 @@
+/********************************************************************************
```
erikbosch, line 1: From a long term perspective I do not see it as optimal that we copy proto files so they exist in multiple repositories (but ok for now). But what would be a better approach? Having the proto-files in separate repository so that both KUKSA.val and kuksa.val.feeders can include it as a submodule, or by some other fetch-mechanism? Or a setup script that feteches the right version from kuksa.val. Any ideas @SebastianSchildt?

someip2val/proto/kuksa_data_broker/kuksa/val/v1/types.proto
```diff
@@ -0,0 +1,286 @@
...
+ * Copyright (c) 2022 Contributors to the Eclipse Foundation
+ *
+ * See the NOTICE file(s) distributed with this work for additional
```
erikbosch, line 4: We do not have a NOTICE file in kuksa.val.feeders, right?

someip2val/src/lib/broker_feeder/actuator_subscriber.cc
```diff
@@ -0,0 +1,173 @@
...
+            LOG_ERROR << "Disconnected with status: "
+                    << subscriber_context_->debug_error_string() << std::endl;
+            if (subscriber_active_) {
+                client_->handleGrpcError(status, "ActuatorSubscriber::Run()");
+                // prevent busy polling if subscribe failed with error
+                std::this_thread::sleep_for(std::chrono::seconds(5));
+            }
+        }
+        {
+            // TODO: lock mutex
```
erikbosch, line 157: When do we need to fix the TODO. Consequences if we do not fix it?

int0x27: it was actually done :) I'll remove the comment

someip2val/proto/kuksa_data_broker/kuksa/val/v1/types.proto
int0x27: protos are all over the place, some with symlinks, I may try fetching them with CMake, but not in this PR

someip2val/src/lib/broker_feeder/collector_client.cc
int0x27: kuksa:: proto was introduced just for the target actuator subscribe as it didn't get merged in sdv:: proto.

sdv:: interfaces support streaming updates, that is lacking in new kuksa:: proto.
But currently we just update datapoints, not stream them anyway.

someip2val/src/lib/broker_feeder/CMakeLists.txt
```diff
@@ -35,10 +35,19 @@ find_program(GRPC_CPP_PLUGIN_PROGRAM_ENV
...
+message("--- [broker_feedder] protoc version:")
+execute_process (
+  COMMAND "${PROTOC_PROGRAM_TEMP}" --version
+)
+
 add_library(data_broker_feeder
   STATIC
     data_broker_feeder.cc
     collector_client.cc
+    # logger.cc
```
erikbosch, line 49: Do we plan to add a logger.cc? Or shall we just remove this comment?

someip2val/src/lib/broker_feeder/actuator_subscriber.cc
```diff
@@ -0,0 +1,173 @@
...
+}
+
+ActuatorSubscriber::ActuatorSubscriber(std::shared_ptr<CollectorClient> client) :
+    client_(client),
+    subscriber_active_(false),
+    cb_(nullptr),
+    actuators_()
+{
+    LOGGER_SET_LEVEL_ENV("KUKSA_DEBUG", LEVEL_INF);
+    // subscriber_context_ = client_->createClientContext();
```
erikbosch, line 41: Remove commented code unless we see a need to keep. If so maybe add a comment on why we keep it.

someip2val/src/lib/broker_feeder/CMakeLists.txt
int0x27: I'll remove it, it turned out more complex to do than expected.

someip2val/proto/kuksa_data_broker/kuksa/val/v1/types.proto
int0x27: We have the same NOTICE reference in every header, better fix it in separate PR

lukasmittag (approved):
What I did:
* Ran everything locally with databroker in debug mode and saw permissions 
```
2023-05-23T14:36:36.867608Z DEBUG databroker::grpc::kuksa_val_v1::val: permissions=Permissions { expires_at: None, read: Everything, actuate: Everything, provide: Everything, create: Everything }
```
* Saw Datapoints get set after executing example

someip2val/src/lib/broker_feeder/collector_client.cc
```diff
@@ -11,55 +11,81 @@
...
+        LOG_TRACE << "Setting dapr-app-id: " << dapr_app_id << std::endl;
+
+    }
+    // allow overriding token file from env
+    std::string databroker_token = auth_token_.empty() ? getEnvVar("BROKER_TOKEN") : auth_token_;
+    if (!databroker_token.empty()) {
+        auto header = "Bearer " + databroker_token;
+        grpc_metadata["authorization"] = header;
+        LOG_TRACE << "Adding authorization: " << header << std::endl;
+    }
```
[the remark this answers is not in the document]
int0x27: Could be logged, but definitely not here: this is called very frequently
There is a --help to show you how to specify a token.
A warning can be logged upon startup

SebastianSchildt (approved):
Let's get this merged, and potentially improve smaller things in later PRs 🚗

PR #105, opened 2023-05-30
initial CSV-recorder

eriksven:
Adds a script for recording files according to the format expected by the CSV-provider

### DIFF
--- a/csv_provider/README.md
+++ b/csv_provider/README.md
@@ -1,16 +1,16 @@
 # VSS Provider
-The aim of this script is to provide and replay VSS data into a `kuksa.val` databroker.
-Therefore, it takes a CSV-file, containting pre-defined sequences of VSS signals including timing delays, and provides it to the 'kuksa.val' databroker.
+The aim of this script is to provide and replay VSS data into a `KUKSA.val` databroker.
+Therefore, it takes a CSV-file, containting pre-defined sequences of VSS signals including timing delays, and provides it to the `KUKSA.val` databroker.
 
 ## Usage
 The provider requires an installation of Python in version 3 and can be executed with the following commands:
 
 ```
 pip install -r requirements.txt
-python provider.py
+python3 provider.py
 ```
 
-This assumes a running `kuksa.val` databroker at `127.0.0.1:55555` and a file named `signals.csv` containing the signals to apply. See the section [Arguments](#arguments) for more details on possible arguments and default values.
+This assumes a running `KUKSA.val` databroker at `127.0.0.1:55555` and a file named `signals.csv` containing the signals to apply. See the section [Arguments](#arguments) for more details on possible arguments and default values.
 
 The provider uses the [kuksa_client]() Python implementation which you need to install on your system, e.g., by applying the [requirement.txt](requirements.txt) with pip.
 
@@ -19,9 +19,9 @@ You can start the provider with the following arguments on a command line:
 
 | short argument | long argument | environment variable | description | default value |
 |---- | ---- | ---- |----- | ----|
-|-f| --file | PROVIDER_SIGNALS_FILE | This indicates the CSV-file containing the signals to update in the `kuksa.val` databroker. | signals.csv |
-| -a | --address | KUKSA_DATA_BROKER_ADDR | This indicates the address of `kuksa.val` databroker to connect to. | 127.0.0.1 |
-| -p | --port | KUKSA_DATA_BROKER_PORT | This indicates the port of the `kuksa.val` databroker to connect to. | 55555 |
+|-f| --file | PROVIDER_SIGNALS_FILE | This indicates the CSV-file containing the signals to update in the `KUKSA.val` databroker. | signals.csv |
+| -a | --address | KUKSA_DATA_BROKER_ADDR | This indicates the address of `KUKSA.val` databroker to connect to. | 127.0.0.1 |
+| -p | --port | KUKSA_DATA_BROKER_PORT | This indicates the port of the `KUKSA.val` databroker to connect to. | 55555 |
 | -i | --infinite | PROVIDER_INFINITE | If the flag is set, the provider loops over the file until stopped, otherwise the file gets processed once. | not present/False
 | -l | --log | PROVIDER_LOG_LEVEL | This sets the logging level. Possible values are: DEBUG, INFO, DEBUG, WARNING, ERROR, CRITICAL | WARNING
 |    | --cacertificate | - | Path to root CA. If defined the client will attempt to use a secure connection and identify the server using this certificate. | None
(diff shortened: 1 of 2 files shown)


csv_provider/README.md
```diff
@@ -42,3 +42,22 @@ Each line in the csv file consists of the following elements:
...
 | value | the new value | 48 |
 | delay | Indicates the time in seconds which the provider shall wait after processing this signal. This way one can emulate the change of signals over time. | 0 |
+
+## Recorder
+One way to generate a CSV-file for the CSV-provider is to record it from an Kuksa.val data broker. This way one can reproduce an interaction of different providers with the Kuksa.val data broker. The script in `csv_provider/recorder` allows this recording.
+An example call, only recording the vehicle speed and width would be:
+
+```
+pip install -r requirements.txt
+python3 provider.py
```
erikbosch, line 52: Is is typo that you call provider.py here, and not recorder.py?

csv_provider/README.md
```diff
@@ -42,3 +42,22 @@ Each line in the csv file consists of the following elements:
 | signal | the name of the signal to update | Vehicle.Speed
 | value | the new value | 48 |
 | delay | Indicates the time in seconds which the provider shall wait after processing this signal. This way one can emulate the change of signals over time. | 0 |
+
+## Recorder
+One way to generate a CSV-file for the CSV-provider is to record it from an Kuksa.val data broker. This way one can reproduce an interaction of different providers with the Kuksa.val data broker. The script in `csv_provider/recorder` allows this recording.
```
erikbosch, line 47: Shall it be:

_The script in `csv_provider/recorder.py`_

csv_provider/README.md
```diff
@@ -42,3 +42,22 @@ Each line in the csv file consists of the following elements:
...
+pip install -r requirements.txt
+python3 provider.py
+```
+
+The recorder supports the following paramters:
+
+| short argument | long argument | description | default value |
+|---- | ---- | ----- | ----|
+|-f| --file | This indicates the filename to which to write the VSS-signals. | signalsOut.csv |
+|-s| --signals | A list of signals to record. | signalsOut.csv | |
```
erikbosch, line 60: Default is incorrect? Shall we mention how you specify multiple signals, i.e. one `-s` with multiple names separated by comma?

csv_provider/README.md
eriksven: True, thanks for pointing this out.

csv_provider/README.md
eriksven: I now extend the example call above to show how one can specify multiple signals even without the comma like in `-s Vehicle.Speed Vehicle.Width`. WDYT?

csv_provider/README.md
```diff
@@ -42,3 +42,22 @@ Each line in the csv file consists of the following elements:
...
+
+The recorder supports the following paramters:
+
+| short argument | long argument | description | default value |
+|---- | ---- | ----- | ----|
+|-f| --file | This indicates the filename to which to write the VSS-signals. | signalsOut.csv |
+|-s| --signals | A list of signals to record. | There is no default value, but the argument is required.| |
+| -a | --address | This indicates the address of `kuksa.val` databroker to connect to. | 127.0.0.1 |
+| -p | --port | This indicates the port of the `kuksa.val` databroker to connect to. | 55555 |
+| -l | --log | This sets the logging level. Possible values are: DEBUG, INFO, DEBUG, WARNING, ERROR, CRITICAL | INFO
```
lukasmittag, line 63: two times DEBUG

lukasmittag (approved):
And just one quick comment. If you set the same value multiple times it will only record only once because databroker just acts on it once. Just letting you know :)

csv_provider/README.md
lukasmittag: @eriksven

lukasmittag:
also one conflicting file I see

csv_provider/recorder.py
```diff
@@ -0,0 +1,96 @@
...
+from kuksa_client.grpc import View
+from kuksa_client.grpc import SubscribeEntry
+
+from kuksa_client.grpc import Field
+
+
+def init_argparse() -> argparse.ArgumentParser:
+    '''This inits the argument parser for the CSV-recorder.'''
+    parser = argparse.ArgumentParser(
+        usage="-a [BROKER ADDRESS] -p [BROKER PORT] -f [FILE]",
```
erikbosch, line 34: Should -s and -l be listed on this line as well?

csv_provider/recorder.py
```diff
@@ -0,0 +1,96 @@
...
+                                               'value': entry.value.value,
+                                               'delay': time_gap})
+                    if entry.actuator_target is not None:
+                        signalwriter.writerow({'field': 'target',
+                                               'signal': entry.path,
+                                               'value': entry.actuator_target.value,
+                                               'delay': time_gap})
+    except VSSClientError as error:
+        logging.error("There was a problem in the interaction"
+                      " with the Kuksa.val data broker at %s:%s: %s ",
```
erikbosch, line 93: Replace "data broker" with "Databroker", see https://github.com/eclipse/kuksa.val/wiki/KUKSA.val-Naming-Guidelines

csv_provider/README.md
```diff
@@ -7,7 +7,7 @@ The provider requires an installation of Python in version 3 and can be executed
 
 ```
 pip install -r requirements.txt
-python provider.py
+python3 provider.py
```
erikbosch, line 10: @SebastianSchildt - I have noticed that we are quite inconsistent regarding if we use `python` or `python3` in our examples. We definitively cannot use "python2" but in most distros "python2" isn't available anyway, and it might be so that your `python` command anyway points to a `python3` installation. We also likely do not support older 3.X-versions anyway.

But in short - do we prefer to use `python` or `python3` in examples? If we agree on preferred term I have an idea of aligning the examples so that they all use the same term. A similar discussion exists for pip vs pip3.

lukasmittag: Personally I prefer python3 because I use it my own but no „strong“ opinion

SebastianSchildt: Just saw the MS Teams bot ate my answer, I wrote this

No strong opinion either. Asking me like this I'd prefer just "python", I do have a vague memory that some of the python3 where needed in the past, as the default was Py2 on some distros also used in containers. But that may have been ancient Ubuntu 16.04 times, so I'd be fine to use "pip" and "python", if container builds break we will see it. From user perspective, depending what environment they use, they can fall into a hole either way....

csv_provider/recorder.py
eriksven: I will replace it to align, but personally I prefer 'data broker' since most language checkers ask for this writing and I do not think the term is unique enough to be a proper noun in the context of Eclipse Kuksa.

erikbosch: I have no strong opinion on what we use as long as we are consistent. At the moment we mostly use "Databroker", in for example the [README](https://github.com/eclipse/kuksa.val/blob/master/kuksa_databroker/README.md), but if we would agreed that "Data broker" would be better name I have no problem with that, but then we should change it "Data broker" in all free-text-places, and possibly use "data-broker" or "data_broker" as preferred name where blanks cannot be used, like directories and docker containers.

SebastianSchildt:
lgtm.

PR #106, opened 2023-06-02
Refactor dbc2val for bidirectional use

erikbosch:
Current status:

Known/Intended limitations, not intended to be solved by this PR:

- val2dbc only supported for Databroker
- It is still called "dbc feeder". 

Tested:

- dbc2val mode
- val2dbc mode
- dbc2val+val2dbc mode
- val2dbc when Databroker requires authentication
- with KUKSA.val server (we shall test that you get an error for val2dbc and no regressions for dbc2val)
- Running dbc2val as docker container


rafaeling:
Review done, looks good to me.

PR #108, opened 2023-06-13
Update README.md with conan install 1.*

MathiasDanzeisen:
Changed the installed conan version to same as in docker build conatiner. To avoid the issue that the install conan now defaults to conan 2.* which is to completly compatable with 1.*

### DIFF
--- a/someip2val/README.md
+++ b/someip2val/README.md
@@ -55,7 +55,7 @@ SOME/IP feeder is [COVESA / vsomeip](https://github.com/COVESA/vsomeip/) applica
     **NOTE:** Sometimes latest conan recipe revisions are broken, but the local build succeeds using cached older revision. If build fails on CI local conan cache could be cleared to reproduce the error. Also latest recipes may require newer conan version.
     ``` bash
     rm -rf ~/.conan/data
-    pip3 install -U "conan==1.55"
+    pip3 install "conan==1.*"
     ```
     Last known working revisions are hardcoded in [conanfile.txt](./conanfile.txt) [requires].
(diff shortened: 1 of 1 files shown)


someip2val/README.md
```diff
@@ -50,12 +50,12 @@ SOME/IP feeder is [COVESA / vsomeip](https://github.com/COVESA/vsomeip/) applica
...
     ``` bash
     sudo apt-get install -y python3 python3-pip
-    pip3 install conan
+    pip3 install "conan==1.55"
     ```
     **NOTE:** Sometimes latest conan recipe revisions are broken, but the local build succeeds using cached older revision. If build fails on CI local conan cache could be cleared to reproduce the error. Also latest recipes may require newer conan version.
     ``` bash
     rm -rf ~/.conan/data
-    pip3 install -U conan
+    pip3 install -U "conan==1.55"
```
int0x27, line 58: -U is redundant with fixed version here, better add to Note that  conan must be < 2.0, due to issues with aarch64 builds.
Another option would be:
```
pip3 install -U "conan<2.0"
```
But there is a hardcoded recipe in `conanfile.txt` that may need attention after updating conan version

int0x27 (changes requested):

someip2val/README.md
MathiasDanzeisen: Ok, how about this?
```suggestion
    pip3 install "conan==1.*"
```
the reason for the PR is that I got conan 2 installen then the build script failed (I think "conan info")

int0x27: Looks fine

int0x27 (approved):
looks good

PR #111, opened 2023-06-16
DBC, GPS, DDS, CSV Feeder TLS support

erikbosch:
This is a companion PR to https://github.com/eclipse/kuksa.val/pull/580

It cannot be merged as the other PR first must be merged, then a pypi package created which can be referenced from this PR

(Edit: a Kuksa-client prerelease ahs been created matching the PR at https://pypi.org/project/kuksa-client/0.4.0a3/)

Design principles:

- Support/Documentation for client certificates removed. Even if kuksa-client supports it there is no reason to use it as Databroker/Server cannot be configured to check client certificates.
- In integration tests we explicitly specify that we want an insecure connection. Not strictly needed now, but good if we later change default in Databroker


dbc2val/requirements.txt
```diff
@@ -1,8 +1,8 @@
 #
-# This file is autogenerated by pip-compile with Python 3.8
+# This file is autogenerated by pip-compile with Python 3.10
 # by the following command:
 #
-#    pip-compile requirements.in
+#    pip-compile --pre requirements.in
```
lukasmittag, line 5: this needs to be reverted back once we have an official release of kuksa-client. Do we remember this or better leave it as is?

lukasmittag (approved):
CSV provider works (but it needs python 3.9 or higher not from you but encountered it)
DBC feeder works, but --tls on command line is not working -> if intended fine
DDS provider works fine
GPS feeder works

dbc2val/Readme.md
```diff
@@ -248,6 +248,8 @@ A smaller excerpt from the above sample, with fewer signals.
 | -                     | *KUKSA_ADDRESS*                 | *[general].ip*          | `127.0.0.1`                      | IP address for Server/Databroker |
 | -                     | *KUKSA_PORT*                    | *[general].port*        | `55555`                          | Port for Server/Databroker |
 | *--tls*               | -                               | *[general].tls*         | `False`                          | Shall tls be used for Server/Databroker connection? |
```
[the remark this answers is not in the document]
erikbosch: Good finding.

We should better have a general discussion at some time how we want to handle configuration in general for KUKSA applications. Today it is a mix of command line arguments, config files and environment variables and it is not uniformly handled. Should better be aligned and perhaps simplified. Supporting all 3 variants also means that you need to decide on order of precedence

dbc2val/requirements.txt
erikbosch: This file is automatically generated, but you are correct, as soon as we have a released kuksa-client we shall update *.in/*.txt files. I tried to write something at https://github.com/eclipse/kuksa.val.feeders/wiki/Release-Process

PR #113, opened 2023-06-27
Support input of multiple DBC files

erikbosch:


dbc2val/Readme.md
```diff
@@ -238,7 +238,7 @@ A smaller excerpt from the above sample, with fewer signals.
 | Command Line Argument | Environment Variable            | Config File Property    | Default Value                    | Description         |
 |:----------------------|:--------------------------------|:------------------------|:---------------------------------|---------------------|
 | *--config*            | -                               | -                       | *See below*                      | Configuration file  |
-| *--dbcfile*           | *DBC_FILE*                      | *[can].dbc*             |                                  | DBC file used for parsing CAN traffic  |
+| *--dbcfile*           | *DBC_FILE*                      | *[can].dbc*             |                                  | DBC file(s) used for parsing CAN traffic. You may specify multiple file names separated by comma|
```
lukasmittag, line 241: maybe give an example because *.dbc, *.dbc won't work even if you put quotes arround it

dbc2val/dbcfeederlib/dbcparser.py
```diff
@@ -28,9 +28,16 @@
 class DBCParser:
     def __init__(self, dbcfile: str, use_strict_parsing: bool = True):
 
-        # Read DBC file
-        log.info("Reading DBC file {}".format(dbcfile))
-        self.db = cantools.database.load_file(dbcfile, strict=use_strict_parsing)
+        self.db = None
+        first = True
+        for filename in dbcfile.split(","):
+            if first:
```
lukasmittag, line 34: here you can strip leading and trailing whitespaces. which lets you specify *.dbc, *.dbc. Also a check if two filenames are the same would be good because else you would get a bunch of warnings with overwriting. Not a priority just a thought.

lukasmittag (changes requested):
What I did:
* Ran against databroker with new option --dbcfile *.dbc,*.dbc
* Ran against databroker speciying *.dbc,*.dbc in the config file

lukasmittag (approved):
Looks good. Tested with command line argument and config works but have a minor comment see below

dbc2val/dbcfeederlib/dbcparser.py
```diff
@@ -28,9 +28,22 @@
...
+            if filename in found_names:
+                log.warn("The DBC file {} has already been read".format(filename))
+            else:
+                found_names.add(filename)
+            if first:
+                log.info("Reading DBC file {} as first file".format(filename))
+                self.db = cantools.database.load_file(filename, strict=use_strict_parsing)
+                first = False
+            else:
+                log.info("Adding definitions from {}".format(filename))
```
lukasmittag, line 45: here you need to check for found_names again, did a test and loading two files says it has loaded the same file but does it anyway after

lukasmittag (changes requested):

dbc2val/dbcfeederlib/dbcparser.py
erikbosch: Yes, right now it just gives a warning, it anyway loads it the second time. I can change so it only loads it once

lukasmittag (approved):

PR #114, opened 2023-06-30
Bugfix Mapper not iterable

mikehaller:
j1939reader is using the Mapper as a Iterable, but the functions were missing:

The fix makes the Mapper iterable, same change as in `vss2ddsmapper.py` (see <https://github.com/eclipse/kuksa.val.feeders/blob/main/dds2val/ddsproviderlib/vss2ddsmapper.py#L117>)

```
2023-06-30 06:36:56,452 WARNING dbcfeederlib.j1939reader: Error decoding message [PGN: CCVS]
  File "/workspaces/leda-example-applications-fork/canbus-j1939/kuksa.val.feeders/dbc2val/dbcfeederlib/j1939reader.py", line 76, in on_message
    if k in self.mapper:
TypeError: argument of type 'Mapper' is not iterable
```

### DIFF
--- a/dbc2val/dbcfeederlib/dbc2vssmapper.py
+++ b/dbc2val/dbcfeederlib/dbc2vssmapper.py
@@ -434,3 +434,9 @@ class Mapper:
             if can_mapping.last_dbc_value is not None:
                 res[can_mapping.dbc_name] = can_mapping.last_dbc_value
         return res
+
+    def __contains__(self, key):
+        return key in self.dbc2val_mapping.keys()
+
+    def __getitem__(self, item):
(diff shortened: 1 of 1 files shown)


erikbosch:
This is likely a regression from a refactoring, unfortunately  we do not have any "public" J1939 test files (dump) that can be run as part of regression tests.

This works, an alternative solution would be to use`self.mapper.get_dbc2val_mappings(k)` like in [dbcreader.py](https://github.com/eclipse/kuksa.val.feeders/blob/main/dbc2val/dbcfeederlib/dbcreader.py#L87)

erikbosch (approved):

PR #116, opened 2023-06-30
Introduce mypy and fix linter errors

erikbosch:


SebastianSchildt (approved):
Looking good, code seems to sneak in some extra checks/logging, but works for me on quick test

PR #118, opened 2023-07-03
Align SOMEIP Feeder docker config

### LINKED ISSUE #115: Docker container registry for someip-feeder
The CI approach for generating and uploading containers for someip-feeder differs quite a lot from other components. Like we have `sha*`-tagged versions, we have an `edge`-version and `latest` point to latest merged commit, for other components typically to latest released version. Other components typically have a `main`version. The approach does not seem to be documented.

Proposal:

- Investigate if a different approach is needed (compared to other components)
- If not, agree that long term approach shall be aligned with other components
- If so, agree on when/how to change handling. As long as we do not remove old releases old references will still work, but you might get a different result than expected (e.g. latest/edge not pointing to latest merged commit, but to something older)
- If someip-feeder actually need a different approach - make sure that we document it, preferably with a rationale.

https://github.com/eclipse/kuksa.val.feeders/blob/main/.github/workflows/someip2val_build.yml#L110
https://github.com/eclipse/kuksa.val.feeders/pkgs/container/kuksa.val.feeders%2Fsomeip-feeder

erikbosch:
This intends to align what docker container that gets published, this PR use the same pattern as Databroker and Dbcfeeder

Fixes #115 

Compare with

* https://github.com/eclipse/kuksa.val.feeders/blob/main/.github/workflows/kuksa_dbc_feeder.yml#L40 
* https://github.com/eclipse/kuksa.val/blob/master/.github/workflows/kuksa_databroker_build.yml#L117

### DIFF
--- a/.github/workflows/someip2val_build.yml
+++ b/.github/workflows/someip2val_build.yml
@@ -114,10 +114,8 @@ jobs:
           images: |
             ghcr.io/${{ github.repository }}/someip-feeder
           tags: |
-            type=sha
-            type=ref,event=tag
-            type=edge
-            type=raw,value=latest,enable={{is_default_branch}}
+            type=ref,event=branch
+            type=ref,event=pr
             type=semver,pattern={{version}}
             type=semver,pattern={{major}}.{{minor}}
             type=semver,pattern={{major}}
@@ -136,6 +134,9 @@ jobs:
             ${{ steps.meta.outputs.tags }}
             ttl.sh/kuksa.val.feeders/someip-feeder-${{github.sha}}:8h
           labels: ${{ steps.meta.outputs.labels }}
+          # Provenance to solve that an unknown/unkown image is shown on ghcr.io
+          # Same problem as described in https://github.com/orgs/community/discussions/45969
+          provenance: false
 
       - name: "Build someip2val container and push to ttl.sh"
         if: ${{ needs.checkrights.outputs.have_secrets != 'true' || github.event_name == 'pull_request' }}
@@ -150,3 +151,6 @@ jobs:
           tags: |
             ttl.sh/kuksa.val.feeders/someip-feeder-${{github.sha}}:8h
           labels: ${{ steps.meta.outputs.labels }}
+          # Provenance to solve that an unknown/unkown image is shown on ghcr.io
+          # Same problem as described in https://github.com/orgs/community/discussions/45969
+          provenance: false


.github/workflows/someip2val_build.yml
```diff
@@ -114,10 +114,8 @@ jobs:
           images: |
             ghcr.io/${{ github.repository }}/someip-feeder
           tags: |
-            type=sha
-            type=ref,event=tag
-            type=edge
-            type=raw,value=latest,enable={{is_default_branch}}
+            type=ref,event=branch
```
erikbosch, line 117: @int0x27 @SebastianSchildt

If we change like this we get the same setup as we use for Databroker and DBCFeeder. Please check if this would cause any problems

int0x27: Looks OK, we don't need sha, latest and edge.
Btw I found a regression happenning from: https://github.com/eclipse/kuksa.val.feeders/pkgs/container/kuksa.val.feeders%2Fsomeip-feeder/96935519?tag=sha-0007f8c
It seems docker image contains 3rd unknown architecture

.github/workflows/someip2val_build.yml
```diff
@@ -150,3 +151,6 @@ jobs:
           tags: |
             ttl.sh/kuksa.val.feeders/someip-feeder-${{github.sha}}:8h
           labels: ${{ steps.meta.outputs.labels }}
+          # Provenance to solve that an unknown/unkown image is shown on ghcr.io
+          # Same problem as described in https://github.com/orgs/community/discussions/45969
+          provenance: false
```
erikbosch, line 156: @int0x27 - It turns out the changed buildx-version caused by the update to build-push-action@v4 two months ago caused oci-manifests to be used, rather than docker manifests. That can be solved by adding `provenance: false`.

This only seems to affect someip-feeder, as the others use our own builder and thus does not use `setup-buildx-action`.


Before adding `provenance: false`

```
erik@debian3:~/kuksa.val/proto/kuksa/val/v1$ docker manifest inspect ttl.sh/kuksa.val.feeders/someip-feeder-44e43a1defc75444feaefd60d5480accbe06309a:8h
{
   "schemaVersion": 2,
   "mediaType": "application/vnd.oci.image.index.v1+json",
   "manifests": [
      {
         "mediaType": "application/vnd.oci.image.manifest.v1+json",
         "size": 1620,
         "digest": "sha256:7e1c4745d5d7d279cfa94318cf453e548656d897a6d63c9cafec003f774ca7ab",
         "platform": {
            "architecture": "amd64",
            "os": "linux"
         }
      },
      {
         "mediaType": "application/vnd.oci.image.manifest.v1+json",
         "size": 1620,
         "digest": "sha256:410250a82b58bd5e1485762f90e70cea89ae7d203a89deff22f292de10492c2f",
         "platform": {
            "architecture": "arm64",
            "os": "linux"
         }
      },
      {
         "mediaType": "application/vnd.oci.image.manifest.v1+json",
         "size": 567,
         "digest": "sha256:4d89de90d5b49718be1184231b8bc715f14a53b4f50a67b30c0ac912aea5ea68",
         "platform": {
            "architecture": "unknown",
            "os": "unknown"
         }
      },
      {
         "mediaType": "application/vnd.oci.image.manifest.v1+json",
         "size": 567,
         "digest": "sha256:b511431bdd319f11785cb08eaf89a51d5f66595de4d7bd456c51165480a74c23",
         "platform": {
            "architecture": "unknown",
            "os": "unknown"
         }
      }
   ]
}
```
After adding `provenance: false`

```
erik@debian3:~/kuksa.val/proto/kuksa/val/v1$ docker manifest inspect ttl.sh/kuksa.val.feeders/someip-feeder-e274e81775597a3c11357a3ee40e3e809ce4b4f8:8h
{
   "schemaVersion": 2,
   "mediaType": "application/vnd.docker.distribution.manifest.list.v2+json",
   "manifests": [
      {
         "mediaType": "application/vnd.docker.distribution.manifest.v2+json",
         "size": 1678,
         "digest": "sha256:df84e70de39310b04c1e27fcf8b21ae6f0ae6b3b035ed7da798a66c17125d564",
         "platform": {
            "architecture": "amd64",
            "os": "linux"
         }
      },
      {
         "mediaType": "application/vnd.docker.distribution.manifest.v2+json",
         "size": 1678,
         "digest": "sha256:6b05b6e681892a9b09a16a4d2dbb46e1618ae00f82874e315b04bf96a28f6b78",
         "platform": {
            "architecture": "arm64",
            "os": "linux"
         }
      }
   ]
}
erik@debian3:~/kuksa.val/proto/kuksa/val/v1$ 
```

int0x27 (approved):
Looks good

PR #120, opened 2023-07-04
Fix kuksa-client dependencies

erikbosch:
It seems we use cache when calling pypi, but we want a fresh build

### DIFF
--- a/dds2val/requirements/requirements-test.txt
+++ b/dds2val/requirements/requirements-test.txt
@@ -1,6 +1,7 @@
+# If pre-releases (e.g. X.YaN) are specified you must use --pre when installing
 pytest
 pytest-html
 pytest-cov
 pytest-asyncio
 py
-kuksa-client > 0.3
+kuksa-client >= 0.4a4
--- a/csv_provider/Dockerfile
+++ b/csv_provider/Dockerfile
@@ -29,7 +29,7 @@ RUN python3 -m venv /opt/venv
 ENV PATH="/opt/venv/bin:$PATH"
 
 RUN /opt/venv/bin/python3 -m pip install --upgrade pip \
-    && pip3 install --no-cache-dir -r requirements.txt
+    && pip3 install --no-cache-dir --pre -r requirements.txt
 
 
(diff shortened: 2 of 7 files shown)


.github/workflows/kuksa_csv_provider.yml
```diff
@@ -95,6 +95,7 @@ jobs:
           linux/arm64
         file: ./csv_provider/Dockerfile
         context: ./csv_provider/
+        no_cache: true
```
sophokles73, line 98: the Action's build output seems to suggest that this needs to be `no-cache: true` instead ...

sophokles73:
Just for my understanding: you now expect the building of the container to fail, because it will not be able to resolve the (alpha) 4.0.x version, right?

erikbosch:
@sophokles73 - When updating the kuksa-client dependencies earlier I made a mistake, I thought `> 0.3` would not match `0.3.1` but that seems to be the case. Not a problem if you use `--pre`, especially not in CI/Docker builds where you start from scratch. So this PR now updates minimum version to latest kuksa-client pre-release, and adds `--pre`for csv provider.

Separately I tried to improve https://github.com/eclipse/kuksa.val.feeders/wiki/Release-Process. When we do a release we need to remove `--pre` and change version, when releasing the feeders possibly even choose a fixed version to make it possible to rebuild using exactly the same version.

dds2val/requirements/requirements-test.txt
```diff
@@ -3,4 +3,4 @@ pytest-html
 pytest-cov
 pytest-asyncio
 py
-kuksa-client > 0.3
+kuksa-client > 0.4a4
```
sophokles73, line 6: shouldn't this also be `>=`, assuming that `0.4a4` is the most recent pre-release?

erikbosch:
@lukasmittag @sophokles73 - please check if you think this one is OK for merge now

lukasmittag (approved):
Looks good!

sophokles73:
I have started the ephemeral image and did not run into the original problem anymore.
LGTM

PR #121, opened 2023-07-05
Support loading CAN message definition files of arbitrary types

sophokles73:
Changed mechanism for adding message definitions from additional files to support arbitrary file formats.

### DIFF
--- a/dbc2val/dbcfeederlib/dbcparser.py
+++ b/dbc2val/dbcfeederlib/dbcparser.py
@@ -20,7 +20,8 @@
 
 import logging
 import sys
-from typing import Set, Optional, Dict, cast
+import os
+from typing import Set, Optional, Dict, cast, Tuple
 import cantools.database
 
 log = logging.getLogger(__name__)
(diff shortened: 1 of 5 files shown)


dbc2val/dbcfeederlib/dbcparser.py
```diff
@@ -34,16 +34,20 @@ def __init__(self, dbcfile: str, use_strict_parsing: bool = True):
         for name in dbcfile.split(","):
             filename = name.strip()
             if filename in found_names:
-                log.warn("The DBC file {} has already been read, ignoring it!".format(filename))
+                log.warning("The DBC file {} has already been read, ignoring it!".format(filename))
                 continue
             found_names.add(filename)
             if first:
```
SebastianSchildt, line 40: Cooudl this logic with "if first" etc. be simplified by creating a database and just adding to it. There seems to be a function `add_dbc(fp)` (or add_dbc_file) and the documentation has this example
```
>>> db = cantools.database.Database()
>>> with open ('foo.dbc', 'r') as fin:
...     db.add_dbc(fin)
```

That would not need a temperary db, that is serialized and then parsed again
I see this probably kills the "automagically determine database type", but there are also e.g. add_kcd functions, if we would try determine ourselves what is the correct one tho choose

Anyway, this is a bit of "armchair coding", I did not test, the code here just did not strike me as super elegant and efficient, and especially on slower targets we found DBC parsing takes a while....

sophokles73: That is possible, of course. The *database* module has a (private) function for determining the database file type which we could copy. If you are that concerned about performance, I can change it accordingly. 

What is the actual use case for loading multiple database files anyway? Is the idea to have one (big) database file that contains the *standard* messages and you then apply some *diffs* via additional files?

sophokles73: Anyway, I have pushed a corresponding change ...

SebastianSchildt: We have setups, where we need to interact with separate components/ECUs on a single CAN Bus, but where DBC description are delivered "per component". Of course it is always possible to merge DBC files via some editing tools, or just concatenating the files, but it is easier/more clean to just be able to load the dbcs you get "as is"

erikbosch (approved):

erikbosch:
> A general question: I see a lot of logging statements that do not use the logging functions' lazy initialization but instead use String formatting to create the log message which is then passed to the logging function. Is that by intention? The linter also seems to be _concerned_ about this and thus I wonder if I am missing something ...

We discussed this at some time - if the use of f-strings like `f"Received value {value}"` is something we would tolerate or not. At that time the conclusion was that it likely isn't a performance problem so we can continue to use it. I have no problem with "forbidding" it. I guess that in most cases it is easy to refactor the log-statements, but that some work might be needed in other cases as maybe not everything written inside a `f"{...}"` easily could be replaced by a `%s` or similar.

(@SebastianSchildt - this is maybe a topic that we shall revisit as part of moving repos to our new orga. Do we need a common repo "kuksa-common" where we specify style/deign-decisions that shall be shared, like max line length, if f-strings in logs should be allowed and so on).

sophokles73:
> I have no problem with "forbidding" it. I guess that in most cases it is easy to refactor the log-statements, but that some work might be needed in other cases as maybe not everything written inside a f"{...}" easily could be replaced by a %s or similar.

If we agree that the other approach is the more desirable one, then we could refactor all occurrences at least in files/places that we are touching in the context of a PR ...

PR #123, opened 2023-07-13
Use default (24 hours) lifetime for ttl.sh images

erikbosch:
Related to https://github.com/eclipse/kuksa.val/issues/598


int0x27 (approved):
Looks good

PR #125, opened 2023-07-24
Improve printout for regsitered signal when using KUKSA.val Server

erikbosch:
No reason to give verbose information in INFO mode 
Replacing f-strings with other construct
(reason: https://pylint.readthedocs.io/en/latest/user_guide/messages/warning/logging-fstring-interpolation.html)


erikbosch:
The failed build might be releated to that we use `--pre` to get the kuksa-client pre-release, but at the same time we get others as well. Build on "main" works, but there it seems cached build is used. Does not seem to be related to changes in this PR. (the changed req files are not part of the original PR, rather added as an attempt to see if an updated files solves the problem)

erikbosch:
The changed Python files can be reviewed. The changed requirement files will anyway be regenerated without `--pre` as soon as we have created kuksa-client 0.4.0

PR #128, opened 2023-07-27
Github CI cleanup as part of 0.4.0 release prep

erikbosch:
Remove some useless builds
Make sure they trigger the same way
Note: DDS is not published to ghcr.io, not changed


.github/workflows/someip2val_build.yml
```diff
@@ -19,6 +19,7 @@ on:
   pull_request:
     paths:
       - ".github/workflows/someip2val_build.yml"
+      - ".github/workflows/check_push_rights.yml"
       - "someip2val/**"
```
lukasmittag, line 23: what about - ".github/actions/post-container-location/action.yml"? included in the other ones but here not

erikbosch: Yes, at the moment someip2val does not use that action, so no need to depend on it. We could potentially try to align how we use thata ction when/if we add ddds2val docker push to ttl/ghcr as part of CI

lukasmittag: Okay, just wanted to confirm that it is wanted :)

lukasmittag (approved):

PR #130, opened 2023-07-28
Make sure we can push tagged builds to ghcr

erikbosch:
This is for manually triggered builds on tags


lukasmittag (approved):
seems right to me. Would we want to update this also in https://github.com/eclipse/kuksa.val/blob/master/.github/workflows/check_push_rights.yml ?

Issue #122, opened 2023-07-06
Investigate if signal names may be reused in DBC files


erikbosch:
A CAN databse consist of a number of messages, each containing a number of signals. At the moment the DBC Feeder design is based on the assumption that signal names are unique. We should better investigate if that is true, or be prepared that it maybe is not always true. According to https://support.vector.com/kb?id=kb_article_view&sysparm_article=KB0011320&sys_kb_id=4ae40ea41b2614148e9a535c2e4bcba7&spa=1 it seems names may be reused in different frames.

A few possible solutions:

**Alt 1: Treat it as a limitation**

State that behavior if there are multiple signals with the same name is undefined. We could refactor `get_canid_for_signal` so that it the first time it search for a signal name does not stop when finding first match, but rather continue and give a warning or error if it finds a second instance. So if you do not get a warning or error you are fine.

**Alt 2: Mandatory message name**

Make message name mandatory in the mapping, make sure it matches.

**Alt 3: Optional message name**

Only require message name if there are conflicts. I.e. if there are 2+ signals with same name give an error if no message name has been specified. if a message name has been specified verify that it match

erikbosch:
FYI @sophokles73 @SebastianSchildt  I would prefer Alt3

sophokles73:
I have conducted some tests with a database file that contains two messages defining a signal of the same name.
The dbparser correctly loads the database and you can look up both messages by name which then both contain a signal of then given name.

I agree that Alt 3 is the least intrusive approach. However, based on the assumption that a signal is only correctly identified by the message name + signal name, I would propose to define this as the default way of specifying a signal and to consistently use this way in all examples. We could then mention that for backwards compatibility reasons it is possible to omit the message name if the signal name is unique across the database.

SebastianSchildt:
**Alt 3** is my vote as well.

As for modifying the example mapping I am on the fence, as I guess due to laziness/usability and the fact most people will map from one small(ish) DBCs, probably the most common, easy pattern is just using signal names.

So what I am saying is: Definitely describe the implications of only using signal name, but not complain/threat too much wrt to warnings during runtime. Perfect would be, if on startup after loading dbc and mapping we can make a check across the database only for uniqueness of signals used in mapping, and if we find an ambiguous case terminate or warn that behaviour is undefined.

sophokles73:
I see (at least) the following use cases:
1. Use a generic DBC file that contains multiple (identical) message definitions for different frame IDs.
   
[...]

PR #132, opened 2023-08-28
Warn on DBC signal name usage across multiple messages

sophokles73:
During startup, the DBC Feeder reads in CAN message definitions
from arbitrary database files. When using the same signal name in
different message definitions, care needs to be taken that the semantics
of the signal are the same across all messages.

The DBC Feeder has been changed to detect such cases and issue a
corresponding warning if the signal is being mapped to a VSS datapoint.

This addresses #122


dbc2val/dbcfeeder.py
```diff
@@ -131,18 +132,18 @@ def __init__(self, client_wrapper: clientwrapper.ClientWrapper,
...
-        dbcfile,
-        mappingfile,
-        dbc_default_file,
-        candumpfile=None,
-        use_j1939=False,
-        use_strict_parsing=False
+        dbcfile: str,
+        mappingfile: str,
+        dbc_default_file: str,
+        candumpfile: str = None,
```
erikbosch, line 138: Mypy check fails in CI, here I believe that it want `optional[str]` as typehint

```
 dbcfeeder.py:138: error: Incompatible default for argument "candumpfile" (default has type "None", argument has type "str")  [assignment]
dbcfeeder.py:138: note: PEP 484 prohibits implicit Optional. Accordingly, mypy has changed its default to no_implicit_optional=True
```

erikbosch:
Conflicts now as well

erikbosch:
@sophokles73 - I think it would be bad if we would loose this PR. Do you intend to work something on it within the next months, or should I try to do some work to make it up-to-date for the new repo?

sophokles73:
@erikbosch let me see if I find the time to migrate the PR to the new repo ...

sophokles73:
@erikbosch I have rebased this PR locally. Once #122 has been migrated to the new repo, I will recreate the PR in the new repo ...

erikbosch:
I created a new issue in the new repo, see https://github.com/eclipse-kuksa/kuksa-can-provider/issues/8

PR #135, opened 2023-09-05
Change dbc2val TLS default to True

### LINKED ISSUE #133: DBC Feeder uses TLS for connection to kuksa.val Server by default
The [README](https://github.com/eclipse/kuksa.val.feeders/blob/3a72e0b7230927205507e5fd96c5162cc26af56a/dbc2val/README.md?plain=1#L251) of the dbc2val feeder suggests that the feeder will not use TLS for connecting to kuksa.val Server/Databroker.

However, [ServerClientWrapper](https://github.com/eclipse/kuksa.val.feeders/blob/3a72e0b7230927205507e5fd96c5162cc26af56a/dbc2val/dbcfeederlib/serverclientwrapper.py#L30) actually seems to default to using TLS.

We should either adapt the documentation or the implementation accordingly.

erikbosch:
Practical differences:

- If use default config for your Databroker instance (no TLS) then you cannot use dbcfeeder defult config (TLS), i.e. just must either change databroker config to use TLS or change dbcfeeder config to not use TLS

Fixes #133 

Note: No other changes to support for example TLS as command line parameter, there we need a general discussion first

### DIFF
--- a/dbc2val/config/dbc_feeder.ini
+++ b/dbc2val/config/dbc_feeder.ini
@@ -29,9 +29,10 @@ ip = 127.0.0.1
 port = 55555
 # port = 8090
 
-# Shall TLS be used (default False for Databroker, True for KUKSA.val Server)
-tls = False
-# tls = True
+# Shall TLS be used
+# Currently KUKSA Server use TLS by default, but KUKSA Databroker does not require TLS by default
+tls = True
+# tls = False
 
 # TLS-related settings
 # Path to root CA, needed if using TLS
--- a/dbc2val/dbcfeeder.py
+++ b/dbc2val/dbcfeeder.py
@@ -436,7 +436,7 @@ def _get_kuksa_val_client(command_line_parser: argparse.Namespace,
         client.set_port(config.getint(CONFIG_SECTION_GENERAL, CONFIG_OPTION_PORT))
 
     if config.has_option(CONFIG_SECTION_GENERAL, CONFIG_OPTION_TLS_ENABLED):
-        client.set_tls(config.getboolean(CONFIG_SECTION_GENERAL, CONFIG_OPTION_TLS_ENABLED, fallback=False))
+        client.set_tls(config.getboolean(CONFIG_SECTION_GENERAL, CONFIG_OPTION_TLS_ENABLED, fallback=True))
 
     if config.has_option(CONFIG_SECTION_GENERAL, CONFIG_OPTION_ROOT_CA_PATH):
         path = config.get(CONFIG_SECTION_GENERAL, CONFIG_OPTION_ROOT_CA_PATH)
--- a/dbc2val/README.md
+++ b/dbc2val/README.md
@@ -248,7 +248,7 @@ A smaller excerpt from the above sample, with fewer signals.
 | *--server-type*       | *SERVER_TYPE*                   | *[general].server_type* | `kuksa_databroker`               | Which type of server the feeder should connect to (`kuksa_val_server` or `kuksa_databroker`) |
 | -                     | *KUKSA_ADDRESS*                 | *[general].ip*          | `127.0.0.1`                      | IP address for Server/Databroker |
 | -                     | *KUKSA_PORT*                    | *[general].port*        | `55555`                          | Port for Server/Databroker |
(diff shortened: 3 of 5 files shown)


dbc2val/dbcfeeder.py
```diff
@@ -436,7 +436,7 @@ def _get_kuksa_val_client(command_line_parser: argparse.Namespace,
         client.set_port(config.getint(CONFIG_SECTION_GENERAL, CONFIG_OPTION_PORT))
 
     if config.has_option(CONFIG_SECTION_GENERAL, CONFIG_OPTION_TLS_ENABLED):
-        client.set_tls(config.getboolean(CONFIG_SECTION_GENERAL, CONFIG_OPTION_TLS_ENABLED, fallback=False))
+        client.set_tls(config.getboolean(CONFIG_SECTION_GENERAL, CONFIG_OPTION_TLS_ENABLED, fallback=True))
```
erikbosch, line 439: I assume it make sense to have the most strict alternative as fallback, so if you for instance write `Ja` or `Nein` it shall be interpreted as `True`

dbc2val/config/dbc_feeder.ini
```diff
@@ -29,9 +29,10 @@ ip = 127.0.0.1
 port = 55555
 # port = 8090
 
-# Shall TLS be used (default False for Databroker, True for KUKSA.val Server)
-tls = False
-# tls = True
+# Shall TLS be used
+# Currently KUKSA Server use TLS by default, but KUKSA Databroker does not require TLS by default
+tls = True
+# tls = False
```
sophokles73, line 35: I guess this line can be removed then ...

argerus:
> Practical differences:
> If use default config for your Databroker instance (no TLS) then you cannot use dbcfeeder defult config (TLS), i.e. just must either change databroker config to use TLS or change dbcfeeder config to not use TLS

What's the rational here... It seems like a step backwards in terms of ease of use.

A more reasonable default would be for the client to connect to the server regardless of whether it's configured with TLS or not, but _fail_ if it can't authenticate the server (when it's using TLS).

That would work seemless with _both_ a default configuration of databroker (no TLS) and with databroker (using TLS) with a proper public key infrastructure in place.

erikbosch:
The "decision" (or maybe just a conclusion) to use TLS as default also for GRPC/Databroker comes from #133, where both @SebastianSchildt and @sophokles73 seems to agree that TLS (only) shall be default for both gRPC (broker) and ws (Server).

As of today at least the gRPC connection fails if you use `tls = False` on client side but the server use TLS, i.e. `grpc.insecure_channel` will fail if the Broker offer only a TLS connection. So to be able to accept both secure and insecure connection if TLS=false would require some changes, for example first test secure and if it fails test insecure (unless tls=true is given)

Concerning default certificates and tokens. For Server the behavior has for a long time been that TLS by default is used/required, and both Server and Client builds (native, docker, PyPI) include default certs and tokens. Previously one reason not to change this would be to avoid breaking backward compatibility for downstream projects. Now when we have released 0.4 that argument may not be that relevant any longer, and I have personally no objection if we would start removing them. But if so one should discuss if default for Broker also shall change. I would personally very much like if we have the same default behavior for both Server and Broker.

@SebastianSchildt  - I think we need some decisions/guidelines from your side on what to do.

SebastianSchildt:
I need to think about it, or maybe we need some more specific proposals to discuss on Dev meeting. Not matter what the "defaults" and options on the various components may turn out to be, I feel very much opposed to use any kind auf "automagic" functionality somewhere, as that may lead to unexpected downgrades. 

If default is insecure, and you need to enable TLS -> fine, if default is secure and you need to disable - > also fine.

But I don't think we want any client/compoent, that I start today and it goes
 - "Let me try TLS, as that is best"
 - "Ah works. Awesome

Whereas the _same_ code/config tomorrow is doing something like
 -  "Let my try  TLS, as I know that is best"
 - "Aw fiddlestticks, doesn't work at all... let me try plain... nobody explicitly told me not to...."
 - "Awesome works!"

That I think is unexpected and may lead to deployment disaster.

erikbosch:
Closing this one for now

PR #137, opened 2023-09-07
Add cancel-in-progress

erikbosch:
Related to https://github.com/eclipse/kuksa.val/issues/626


SebastianSchildt (approved):
Same argument as https://github.com/eclipse/kuksa.val/pull/651 

Let's get this merge, sit back and observe

PR #138, opened 2023-09-07
Merge DBCParser into Mapper

sophokles73:
The readers require a Mapper as well as information provided by the DBCParser. The Mapper itself also requires information from the parser.

The Mapper has therefore been changed to use the DBCParser class as a mixin. The readers and the Feeder have been adapted accordingly.


dbc2val/dbcfeederlib/dbc2vssmapper.py
```diff
@@ -1,7 +1,7 @@
 #!/usr/bin/python3
 
 ########################################################################
-# Copyright (c) 2020 Robert Bosch GmbH
+# Copyright (c) 2020,2023 Robert Bosch GmbH
```
erikbosch, line 4: @SebastianSchildt - do we somewhere have any guidelines on how to handle "years" in copyright? As of today we have for dbcfeeder two types of license statements, one (like here) stating Bosch copyright, others have a more generic "Contributors to the Eclipse Foundation". Then for year, when shall we add a new year like here. Always?

sophokles73: The same section also states for the newer copyright header's 1st line:

> The year is expressed as a single year or a range spanning the year of the very initial assertion and that of the most recent change.

:-)

dbc2val/test/test_j1939/test_j1939.py
```diff
@@ -1,72 +0,0 @@
-#!/usr/bin/python3
-
-########################################################################
-# Copyright (c) 2023 Contributors to the Eclipse Foundation
```
erikbosch, line 4: Intentional to remove this file? - it was just added, right?

sophokles73: it's the latter :-)

erikbosch (approved):

PR #141, opened 2023-09-11
Fix bit-mask used for filtering CAN frames in elmbridge

sophokles73:


erikbosch (approved):

erikbosch:
Looks good to me, I like that you added some comments that explains the code

PR #143, opened 2023-09-18
Introduce CAN reader hierarchy

sophokles73:
Increase cohesion and reduce coupling between components by introducing
an abstract CANReader class which defines the responsibilities and
behavior of the DBCReader and J1939Reader.

Modified readers to consistently use HW based CAN frame filtering based
on DBC message and DBC<->VSS mapping definitions.

Also added typing information to functions and member variables to
improve readability and maintainability.


erikbosch (approved):
I tested this PR both for local build and docker build and works without problem. During testing I detected a totally unrelated problem which is addressed in #144

PR #145, opened 2023-10-03
Using global check push right

erikbosch:
Corresponds to https://github.com/eclipse/kuksa.val/pull/680


erikbosch:
Candidate for v0.4.1 - can you review @SebastianSchildt @lukasmittag

SebastianSchildt (approved):
Changes and CI runs look good. If something is messed up, we weill notive after tagging....

PR #147, opened 2023-10-23
Ignore signal values that are out of the defined range

sophokles73:
The message definition database may contain signals for which a
minimum and/or maximum value has been defined. In particular,
j1939 uses "special" out-of-range values to indicate that a signal
is not available (at the moment). These "special" values should not
be published to the Databroker.

The CanReader has been altered to ignore signal values from CAN
messages which do not match the defined min/max values.


erikbosch (approved):
Looks good to me, smoke test performed

PR #157, opened 2023-12-14
Remove self-hosted, remove alpine

erikbosch:
Refactoring builds, trying to achieve:

- Remove dependency to self-hosted builder, as the future of that builder is uncertain
- Remove alpine, as that typically leads to very long ghcr build times
- Accept somewhat bigger image size

This should fix the failing builds we are currently experiencing for arm64 for self-hosted.
I have done some basic tests locally.
(This also enables arm builds for gpsfeeder, was previously disabled)

### DIFF
--- a/.github/workflows/kuksa_gps_feeder.yml
+++ b/.github/workflows/kuksa_gps_feeder.yml
@@ -23,103 +23,17 @@ jobs:
     uses: eclipse-kuksa/kuksa-actions/.github/workflows/check_ghcr_push.yml@2
     secrets: inherit
 
-  build-self-hosted:
-    if: ${{ github.repository_owner == 'eclipse' }}
-    runs-on: [ self-hosted ]
-    needs: check_ghcr_push
 
-    # With default permissions, release action fails on forks
-    permissions:
-      contents: read
-      packages: write
-
-    steps:
-    - uses: actions/checkout@v3
-
-    - name: Docker meta
-      id: meta
-      uses: docker/metadata-action@v4
-      with:
-        # list of Docker images to use as base name for tags
-        images: ghcr.io/${{ github.repository }}/gps
-        # generate Docker tags based on the following events/attributes
-        tags: |
-          type=ref,event=branch
-          type=ref,event=pr
-          type=semver,pattern={{version}}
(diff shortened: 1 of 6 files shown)


.github/workflows/kuksa_gps_feeder.yml
```diff
@@ -70,119 +72,30 @@ jobs:
         push: true
         tags: |
           ${{ steps.meta.outputs.tags }}
+          ${{ env.EPHEMERAL_IMAGE }}
```
lukasmittag, line 75: why do we add this?

lukasmittag: but there's no tag ```${{ env.EPHEMERAL_IMAGE }}``` in them or am I missing something?

erikbosch: In [gps feeder](https://github.com/eclipse/kuksa.val.feeders/blob/main/.github/workflows/kuksa_gps_feeder.yml#L18) someone has by some reason defined an env variable:


```
env:
  EPHEMERAL_IMAGE: "ttl.sh/kuksa.val.feeders/gps-${{ github.sha }}"
```

which in this PR is used as

```
        tags: |
          ${{ steps.meta.outputs.tags }}
          ${{ env.EPHEMERAL_IMAGE }}
```

In the other repos no environment variable exists, instead the ttl.sh name is stated explicitly.



```
        tags: |
          ${{ steps.meta.outputs.tags }}
          ttl.sh/kuksa.val/kuksa-databroker-${{github.sha}}
```

```
        tags: |
          ${{ steps.meta.outputs.tags }}
          ttl.sh/kuksa.val/kuksa-dbcprovider-${{github.sha}}
```

So the end result shall be roughly the same, if you build on main/master/tag you will upload to both ttl.sh and ghcr.io. I however just noticed that the naming pattern differs for ttl.sh images, like if you compare `ttl.sh/kuksa.val/kuksa-dbcprovider-${{github.sha}}` with `ttl.sh/kuksa.val.feeders/gps-${{ github.sha }}`. But as we are about to move the feeders anyway that possibly does not make sense to fix.

lukasmittag (approved):
LGTM

PR #161, opened 2023-12-24
Introducing eCAL Feeder for KUKSA.val Integration

lalywr2000:
### Background

This pull request contains features implemented as part of the 2023 Eclipse SDV Hackathon Challenge.

### Description

This pull request introduces a new utilization example for KUKSA.val through the implementation of the eCAL Feeder. The primary objective of this implementation is to facilitate the input of data received via eCAL into the KUKSA.val databroker. The communication between eCAL and KUKSA.val is established through protobuf-formatted topics, utilizing the VSS (Vehicle Signal Specification) description and data structure outlined in this format. The data is seamlessly provided to the KUKSA.val databroker via the kuksa_client.

[About eclipse eCAL](https://eclipse-ecal.github.io/ecal/)


lukasmittag:
Hey @lalywr2000, thanks for the Pull Request. In order to get the CI running you need to create an eclipse account https://accounts.eclipse.org/user/register and sign the eclipse eca https://www.eclipse.org/legal/ECA.php https://www.eclipse.org/legal/ecafaq.php

lalywr2000:
Thanks for the instructions. @lukasmittag Now I have thet ECA Validation.

lukasmittag:
Hey @lalywr2000, I see CI is failing due to no SPDX header in the proto file can you fix this?

ecal2val/mock_publisher.py
```diff
@@ -0,0 +1,55 @@
...
+
+            protobuf_message = vss_data_pb2.VssData()
+            protobuf_message.description = description
+            protobuf_message.data_int = 0
+            protobuf_message.data_float = 0
+
+            try:
+                protobuf_message.data_int = int(data)
+            except ValueError:
+                protobuf_message.data_float = float(data)
```
lukasmittag, line 48: what if it's a string/bool etc.?

ecal2val/proto_struct/vss_data.proto
```diff
@@ -0,0 +1,10 @@
+syntax = "proto3";
+
+package proto_struct;
+
+message VssData
+{
+  string description = 1;
+  int32  data_int    = 2;
```
lukasmittag, line 22: Couldn't you leverage the structure of https://github.com/eclipse/kuksa.val/blob/master/proto/kuksa/val/v1/types.proto so for example use Datapoint etc. directly and then you not need to handle that much for float/bool/string etc.

lalywr2000:
@lukasmittag To leverage the structure of 'types.proto' effectively, I understood that I should use 'kuksa_client.grpc' library. However, I encountered difficulties in using it directly for topic communication in eCAL.

Therefore, the current completed code looks like this. A custom-made proto structure based on 'types.proto' is utilized in communication between eCAL. And for writing data to the databroker, 'kuksa_client.grpc' is used normally.

SebastianSchildt:
Some Linter failures, you may want to check the notes on pre-commit https://github.com/eclipse/kuksa.val.feeders#pre-commit-set-up for a way to check this locally already. It is annoying the first we times it bites you, but then it becomes natural 🚀

lalywr2000:
@SebastianSchildt Thank you. Finished installing pre-commit and solved every issues locally.

SebastianSchildt (approved):
Thanks again for the contribution. Sorry this took far too long, first forgot it, then could not find a x64 machine which ecal seems to require

Anyway: Followed the steps in Readme and "works as advertised", checked reception of data with kusksa-client

@erikbosch  I think for stuff like this,we really need an "experiments"/"incubation" repo when migrating: We want it, it is cool, but currently is PoC level: Useful but only time will tell it will become "first class provider" such as can.

PR #165, opened 2024-02-02
Removing GPS Provider

erikbosch:
This is the last step of moving GPS feeder to the new repo https://github.com/eclipse-kuksa/kuksa-gps-provider. A last gps feeder docker container with a warning has been published to https://github.com/eclipse/kuksa.val.feeders/pkgs/container/kuksa.val.feeders%2Fgps, so no more steps remain as I see it.


lukasmittag (approved):
LGTM

PR #167, opened 2024-02-09
Removing CSV and SOMEIP Provider

erikbosch:
Migration is now finished


SebastianSchildt (approved):
Verified we are not deleting anything of use, and checked that published images print a notice.

lgtm ↗️

PR #168, opened 2024-02-23
Deprecate replay

### LINKED ISSUE #69: Replay not working anymore
The replay function only works for KUKSA.val server. Maybe it should be dropped? Definitely should be clearly marked outdated.

erikbosch:
Fixes #69 

(*well sort of, or rather indicates that we do not want to fix it)

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,7 +6,8 @@
 
 Name | Description
 ---- | -----------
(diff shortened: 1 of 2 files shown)


SebastianSchildt (approved):
lgtm

